PR #16, opened 2020-12-22
Remove "debugging" printf calls from nx_tcpserver.c

hwmaier:
Those two printf's are probably a left-over from a debugging session and should not be in released code.


TiejunMS:
Thanks @hwmaier ! These changes will be applied in next minor release.

PR #43, opened 2021-08-03
Prevent Race Condition with packet releases in nx_tcp_socket_state_ack_check()

hwmaier:
In certain conditions TCP segment ACKs can arrive earlier than corresponding packet TX complete interrupts. This is in particular true for hardware which only generates an interrupt once a set of TX descriptors is complete, rather for each individual packet.

In such situations I found that if `nx_packet_transmit_release()` is called from a TX complete interrupt which occurs while processing the packet release in `nx_tcp_socket_state_ack_check()`, a race condition can occur due to either instruction caching or branch prediction on Cortex-M MCUs. A packet can wrongly become marked `NX_PACKET_ALLOCATED` and `NX_DRIVER_TX_DONE` at the same time and as a result the packet will never be released, neither by `nx_tcp_socket_state_ack_check()` nor by the TX ISR calling `nx_packet_transmit_release()`.
  Over time such a system will run out of free NX_PACKETs and stop communicating.

  This race condition is very difficult to reproduce as slightest change in timing may prevent the race condition to happen. As such I cannot provide a simple test case.

## Background

`nx_packet_transmit_release()` contains the following code to mark packets with a NX_DRIVER_TX_DONE flag to postpone the packet release after either retransmission or acknowledgement. This code sequence is wrapped in TX_DISABLE/TX_RESTORE to disable interrupts:

```
    if ((packet_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next != ((NX_PACKET *)NX_PACKET_ALLOCATED)) &&
        (packet_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next != ((NX_PACKET *)NX_PACKET_FREE)))
    {


        /* Yes, this is indeed a TCP packet.  Just mark this with the NX_DRIVER_TX_DONE
           value to let the TCP layer know it is no longer queued up.  */
        /*lint -e{923} suppress cast of ULONG to pointer.  */
        packet_ptr -> nx_packet_queue_next =  (NX_PACKET *)NX_DRIVER_TX_DONE;
``` 

`nx_tcp_socket_state_ack_check()` on the other hand contains the following complementary code to release packets which have been sent and acknowledged. Packets which can be released are marked with `NX_PACKET_ALLOCATED`. Packets which have been marked with `NX_DRIVER_TX_DONE` by the ISR can be immediately released, others packets are expected to be released by the ISR after completed transmission.

```
        /* Now walk through the packets to release and set them
           free.  */
        while (packet_release_count--)
        {

            /* Use the previous pointer as the release pointer.  */
            previous_ptr =  search_ptr;

            /* Move to the next packet in the queue before we clip the
               next pointer.  */
            search_ptr =  search_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next;

            /* Set the packet to allocated to indicate it is no longer part of the TCP queue.  */
            /*lint -e{923} suppress cast of ULONG to pointer.  */
            previous_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next =  ((NX_PACKET *)NX_PACKET_ALLOCATED);
            NX_PACKET *transmit_done_flag = previous_ptr -> nx_packet_queue_next;

            /* Has the packet been transmitted? This is only pertinent if a retransmit of
               the packet occurred prior to receiving the ACK. If so, the packet could be
               in an ARP queue or in a driver queue waiting for transmission so we can't
               release it directly at this point.  The driver or the ARP processing will
               release it when finished.  */
            /*lint -e{923} suppress cast of ULONG to pointer.  */
            if (transmit_done_flag == ((NX_PACKET *)NX_DRIVER_TX_DONE))
            {

                /* Yes, the driver has already released the packet.  */
```

## Test Setup and Bug Detection

On a Rensesas EK-RA6M5 evaluation board we run a web server application which is continuously polled using a wget client running on a PC. The web server application also runs a cyclic monitor thread ever 100ms which inspects the packet pool and searches for orphaned NX_PACKETs using the following routine:

```
static void debugPackets(NX_PACKET_POOL *pool_ptr)
{
    for (UINT i = 0; i < pool_ptr->nx_packet_pool_total; i++)
    {
        NX_PACKET *packet;
        ULONG packet_status;
        CHAR *thread_info;
        CHAR *file_info;
        ULONG line;

        _nx_packet_debug_info_get(pool_ptr, i, &packet, &packet_status, &thread_info, &file_info, &line);

        // Orphaned packet?
        if (packet_status == 0xAAAAAAAA && packet->nx_packet_queue_next == (void *)0xDDDDDDDD)
        {
            // Log it
            SEGGER_RTT_printf(0, "*** Found orphanded packet %p!\n", packet);
        }
    }
}
```

Without the fix, the monitor thread will report within a few minutes orphaned packets and eventually the system will run out of free packets.

## Solutions

Both solutions have to be applied to `nx_tcp_socket_state_ack_check()` after line 667.

Option 1: ARM specific, add a __DSB instruction to clear branch predictor and ensure all outstanding memory accesses are completed before evaluating `->nx_packet_queue_next`:

```
            previous_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next =  ((NX_PACKET *)NX_PACKET_ALLOCATED);

            __DSB();  // Data Synchronization Barrier

            if (previous_ptr -> nx_packet_queue_next ==  ((NX_PACKET *)NX_DRIVER_TX_DONE))
            {
```         

Option 2: This solution is more generic and should work also on non-ARM MCUs and wraps the two lines in a TX_DISABLE/TX_RESTORE pair. It is implemented in this PR.

```
            TX_DISABLE

            previous_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next =  ((NX_PACKET *)NX_PACKET_ALLOCATED);
            NX_PACKET *transmit_done_flag = previous_ptr -> nx_packet_queue_next;

            TX_RESTORE

            if (transmit_done_flag == ((NX_PACKET *)NX_DRIVER_TX_DONE))
            {            
```

### DIFF
--- a/common/src/nx_tcp_socket_state_ack_check.c
+++ b/common/src/nx_tcp_socket_state_ack_check.c
@@ -88,6 +88,7 @@
 /**************************************************************************/
 UINT  _nx_tcp_socket_state_ack_check(NX_TCP_SOCKET *socket_ptr, NX_TCP_HEADER *tcp_header_ptr)
 {
+TX_INTERRUPT_SAVE_AREA
 
 NX_TCP_HEADER *search_header_ptr = NX_NULL;
 NX_PACKET     *search_ptr;
@@ -673,9 +674,19 @@ UINT           wrapped_flag = NX_FALSE;
                next pointer.  */
             search_ptr =  search_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next;
 
+            /* Disable interrupts temporarily.
+               Following sequence must be atomic or contain a Data Synchronization Barrier before
+               nx_packet_queue_next is evaluated. Without it a race condition with the TX ISR calling
+               nx_transmit_release() can occur and we end up with packets never being released. */
+            TX_DISABLE
+
             /* Set the packet to allocated to indicate it is no longer part of the TCP queue.  */
             /*lint -e{923} suppress cast of ULONG to pointer.  */
             previous_ptr -> nx_packet_union_next.nx_packet_tcp_queue_next =  ((NX_PACKET *)NX_PACKET_ALLOCATED);
+            NX_PACKET *transmit_done_flag = previous_ptr -> nx_packet_queue_next;
+
+            /* Restore interrupts.  */            
+            TX_RESTORE
 
             /* Has the packet been transmitted? This is only pertinent if a retransmit of
                the packet occurred prior to receiving the ACK. If so, the packet could be
@@ -683,7 +694,7 @@ UINT           wrapped_flag = NX_FALSE;
                release it directly at this point.  The driver or the ARP processing will
                release it when finished.  */
             /*lint -e{923} suppress cast of ULONG to pointer.  */
-            if (previous_ptr -> nx_packet_queue_next ==  ((NX_PACKET *)NX_DRIVER_TX_DONE))
+            if (transmit_done_flag == ((NX_PACKET *)NX_DRIVER_TX_DONE))
             {
 
                 /* Yes, the driver has already released the packet.  */


wenhui-xie:
Thank you very much for catching this issue @hwmaier. And thanks for your suggestions. We will fix it in the future release.

PR #61, opened 2021-12-13
Fix BSD sendto declarations

josesimoes:
- Where missing the const in the msg parameter.


TiejunMS:
@josesimoes , thanks for your PR! We recognize the missing `const` in the `sendto` API. But to keep backward compatibility, we can not change released APIs. Changes to this API will impact existing users who have already adapt their applications with this parameter without `const`.

PR #132, opened 2022-11-07
Add PFC and ACFC support in PPP driver.

ohoc:
Add support for Protocol-Field-Compression (PFC) and Address-and-Control-Field-Compression (ACFC) as defined in RFC 1661.


ohoc:
@microsoft-github-policy-service agree [company="Exelen GmbH"]

bo-ms:
The features of PFC and ACFC are supported now.

PR #197, opened 2023-10-18
dump connection status when nx_azure_iot_hub_client_connect failed

ks6088ts:
Currently the result code cannot be dumped when `nx_azure_iot_hub_client_connect()` failed, which makes our troubleshoot difficult.


ks6088ts:
@TiejunMS Could you take a look at it?

PR #229, opened 2023-12-28
Updated version and date for 6.4.0

bo-ms:


nx_secure/inc/nx_secure_tls.h
```diff
@@ -166,7 +166,7 @@ extern   "C" {
 
 #define AZURE_RTOS_NETX_SECURE
 #define NETX_SECURE_MAJOR_VERSION                       6
-#define NETX_SECURE_MINOR_VERSION                       3
+#define NETX_SECURE_MINOR_VERSION                       4
```
TiejunMS, line 169: Function header needs to be updated

scripts/update_patch.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
TiejunMS, line 1: This file is no longer needed

bo-ms: removed

TiejunMS (approved):

PR #252, opened 2024-03-01
Fix compile error with NX_DEBUG enabled

madsknielsen:
When compiling with NX_DEBUG enabled the
nx_ram_network_driver gives a compiler error when
printing the IP Address, since long sub-specifier
is missing.


fdesbiens:
Thank you for this contribution, @madsknielsen. My apologies for the long delay.

PR #264, opened 2024-03-28
add stm32 network drivers

rahmanih:


wickste:
Thank you @rahmanih - this looks good!

I will need to hold off on merging it for the moment as we are still finalizing the new development process. We need to make sure all new contributions are compliant with that process, so we can continue to obtains safety certification for this code base.

jerrylogansquare:
Hello,
I've been trying to use this code in our project with LAN9354 (3-port switch device).  When I add two IP instances using nx_ip_create, it calls _nx_driver_initialize(NX_IP_DRIVER *driver_req_ptr) twice.  There was no indication it couldn't really create two IP instances due to the limitation of the driver.  The second IP instance simply overwrites the first.  I don't think this is a bug per se, but its not well advertised by NetX Duo that your ethernet driver *must* support multiple IPs also, not just the networking stack.  I'm now looking into creating two instances of the ethernet driver with two different MAC addresses.  
Jerry

rahmanih:
closing the PR as it'll be delivered in a separate repository.

PR #265, opened 2024-03-29
Fix key_size_in_bits shift direction in _nx_crypto_method_hkdf_init

sjscymru:
This change fixes an error in the shift direction when setting the `nx_crypto_hkdf_ikm_length` in the `_nx_crypto_method_hkdf_init` function.

The parameter `key_size_in_bits` should be shifted to the right to set the correct length in bytes.


fdesbiens:
Thank you for this contribution, @sjscymru.

I made some regression test tweaks to match the new code. Merging to dev now. This will ship in our Q2 2026 release.

PR #269, opened 2024-04-10
MQTT / nxd_mqtt_client_message_get() : Release packet if topic or message is bigger than destination buffer

EdouardMALOT:
When configuring the MQTT client to subscribe to a topic, we receive payloads of unknown size.

To read the MQTT message (and release the received TCP packet), we must use nxd_mqtt_client_message_get() with the buffer and buffer size as arguments.

If the received message (payload or topic) is larger than the user buffer, this function returns an error but does not release the TCP packet. This can cause an endless issue where the client cannot receive any more messages (nxd_mqtt_client_message_get() always responds with an error without releasing the TCP packet).

In this pull request, the received packet is released if the destination buffer is too small. This way, messages that are too large are rejected (the user is informed with the return value NXD_MQTT_INSUFFICIENT_BUFFER_SPACE), and the user can continue to receive smaller messages later.

### DIFF
--- a/addons/mqtt/nxd_mqtt_client.c
+++ b/addons/mqtt/nxd_mqtt_client.c
@@ -5438,15 +5438,6 @@ ULONG               message_length;
     {
         packet_ptr = client_ptr -> message_receive_queue_head;
         status = _nxd_mqtt_process_publish_packet(packet_ptr, &topic_offset, &topic_length, &message_offset, &message_length);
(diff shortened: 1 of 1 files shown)


hwmaier:
Hi @EdouardMALOT,
Please refer to my previous PR https://github.com/eclipse-threadx/netxduo/pull/14 which is related.

EdouardMALOT:
This issue is also discussed with another approach here: #14

PR #304, opened 2025-03-04
Fix/internal http put

hnguyenHWI:
Ports the fixes for NetX Duo HTTP server advisories from commit c78d650be7377aae1a8704bc0ce5cc6f9f189014 to the internal HTTP module. 

(TLS does not seem to be used so there is no corresponding ported change of the check for an empty packet, only the HTTP PUT changes are included).


fdesbiens (approved):
Thanks for this contribution!

PR #317, opened 2025-05-15
Fix HKDF implementation to prevent buffer overrun when compiled with NX_SECURE_KEY_CLEAR

sjscymru:
The `_nx_crypto_method_hkdf_cleanup` method is passing the incorrect pointer to the HMAC method cleanup when NetX is compiled with `NX_SECURE_KEY_CLEAR`.

When `NX_SECURE_KEY_CLEAR` defined, calling `_nx_crypto_method_hkdf_cleanup` will result in a subsequent call to the HMAC method (if one is set) cleanup function. However the incorrect parameter is being passed, currently HKDF passes in the a pointer to the `NX_CRYPTO_METHOD` structure of the HMAC method, which is 32 bytes in size.

```
status = hmac_method -> nx_crypto_cleanup(hmac_method);
```

But the HMAC cleanup method is calling `NX_CRYPTO_MEMSET()` on the provided pointer using `sizeof(NX_CRYPTO_HMAC)` which is 292 bytes is size:

```  
/* Clean up the crypto metadata.  */
NX_CRYPTO_MEMSET(crypto_metadata, 0, sizeof(NX_CRYPTO_HMAC));
```

This leads to a buffer overflow of 260 bytes. The actual amount can vary depending of the exact HMAC method being used.

### DIFF
--- a/crypto_libraries/src/nx_crypto_hkdf.c
+++ b/crypto_libraries/src/nx_crypto_hkdf.c
@@ -172,7 +172,7 @@ UINT status;
 
     if(hmac_method)
     {
-        status = hmac_method -> nx_crypto_cleanup(hmac_method);
+        status = hmac_method -> nx_crypto_cleanup(hkdf->nx_crypto_hmac_metadata);
 
         if (status != NX_CRYPTO_SUCCESS)
         {


fdesbiens:
Thank you for this contribution, @sjscymru. 

@eclipse-threadx/iot-threadx-committers: Please review.

PR #372, opened 2026-03-25
ftp server: call logout callback on client timeout for consistency

vb-linetco:
Otherwise this is not symmetric with the other case where logout is also called when the client disconnects.

Refactor performing the logout into a separate helper to avoid duplication.


fdesbiens:
Thank you for this contribution @vb-linetco. I will review this soon.

addons/ftp/nxd_ftp_server.c
```diff
@@ -6498,6 +6445,78 @@ UINT    size;
...
 
+/**************************************************************************/ 
+/*                                                                        */ 
+/*  FUNCTION                                               RELEASE        */ 
+/*                                                                        */ 
+/*    _nx_ftp_server_logout_client                        PORTABLE C      */ 
+/*                                                           6.4.3        */
+/*  AUTHOR                                                                */
+/*                                                                        */
+/*      TBD                                                               */
```
vb-linetco, line 6456: @fdesbiens Since I've only moved this code, I was not sure which Autor to put here. What is the policy about this?

fdesbiens: @vb-linetco My apologies for the late reply. AUTHOR should be the original author. In this case, you could have both the original and yours since you initiated the refactoring.

addons/ftp/nxd_ftp_server.c
```diff
@@ -5187,6 +5161,9 @@ NX_FTP_CLIENT_REQUEST   *client_req_ptr;
                 /* Reset the block bytes.  */
                 client_req_ptr -> nx_ftp_client_request_block_bytes = 0;
 
+                /* Logout the client. */
+                _nx_ftp_server_logout_client(ftp_server_ptr, client_req_ptr);
```
fdesbiens, line 5163: You log out due to an inactivity timeout, but leave `nx_ftp_client_request_authenticated` set. The timed-out request is then disconnected and relistened, while `_nx_ftp_server_connect_process` does not reset authentication for the next accepted client. Since command processing only blocks unauthenticated clients at `addons/ftp/nxd_ftp_server.c:1553`, the next connection can inherit authenticated state and issue commands without USER/PASS. The timeout path should clear authentication as the disconnect path does. See `_nx_ftp_server_control_disconnect_processing` for more details.

addons/ftp/nxd_ftp_server.c
```diff
@@ -1720,39 +1722,9 @@ ULONG                   block_size;
                     /* Increment the number of disconnection requests.  */
                     ftp_server_ptr -> nx_ftp_server_disconnection_requests++;
 
-                    /* Check if this client login.  */
-                    if (client_req_ptr -> nx_ftp_client_request_login)
-                    {
-
-                        /* Call the logout function.  */
```
fdesbiens, line 1727: There is some trailing whitespace here and in the comment block for your new helper. Please make sure to clean that up. Thanks!

fdesbiens (changes requested):
Thanks for this contribution, @vb-linetco. 

I requested some small changes. Moreover, you did not add regression tests for the new timeout behaviour. Please add one that logs in, lets the FTP activity timeout fire, asserts the logout callback count increments, then reconnects and verifies commands are rejected until a fresh login occurs. 

I look forward to merging this useful improvement.

fdesbiens:
I edited the PR to implement my own feedback. Merging to dev now. This will ship with the Q2 2026 release.

vb-linetco:
Thank you for addressing the feedback, I was planning to do it today :+1:

PR #374, opened 2026-04-09
BSD socket options improvements

mzgrebnak:
- support for SO_LINGER option
- aligning nx_bsd_setsockopt with nx_bsd_getsockopt


fdesbiens:

fdesbiens:
Hi @mzgrebnak.

 Thanks for the contribution. I tested this with the repo recloned using submodules, and the BSD-enabled build currently fails.

  Command run:

  ```bash
  ./test/cmake/netxduo/run.sh build v4_full_build
  ```
---
**Blocking issues:**

1. `addons/BSD/nxd_bsd.c` does not compile because the patch reads/writes `NX_TCP_SOCKET::so_linger`, but `NX_TCP_SOCKET` has no `so_linger` member.

     Affected locations include:

     - `addons/BSD/nxd_bsd.c:4894`
     - `addons/BSD/nxd_bsd.c:6291`
     - `addons/BSD/nxd_bsd.c:6738`

    Linger state should likely be stored in the BSD socket layer (`NX_BSD_SOCKET`), not in the NetX TCP socket structure. The BSD-layer socket state is defined separately as NX_BSD_SOCKET in addons/BSD/nxd_bsd.h:936, and it already has linger-related fields around addons/BSD/nxd_bsd.h:963.

2. The new `IP_TOS` setter also fails under `-Werror`:

     - `addons/BSD/nxd_bsd.c:6648`

    The code assigns a signed `int` option value into `ULONG`, then checks `type_of_service >= 0`, which triggers `-Wsign-conversion` and `-Wtype-limits`.

---
**Additional issues to fix after the build failure:**

- The `IPPROTO_IP` option validation now accepts `SOL_SOCKET` option numbers because the lower-bound check was removed. This can make calls with the wrong option level dispatch into unrelated `SO_*` cases.
- `getsockopt()` cases such as `SO_BROADCAST` and `TCP_NODELAY` write an `INT` without validating `*option_length`.
- `IP_TTL` handling treats any non-UDP socket as TCP, which can dereference a null TCP socket pointer for raw sockets.

Please update the implementation and add targeted BSD regression coverage for the new socket options, especially `SO_LINGER`, `IP_TOS`, and `IP_TTL`.

fdesbiens (changes requested):
Please make changes per my comment.

fdesbiens:
@mzgrebnak I ended up updating the PR myself since I wanted to ship this with our Q2 2026 release. Will do the same for #375 and #376.

PR #375, opened 2026-04-09
BSD select improvements

mzgrebnak:
- select improvements
- exception notify added
- nx_bsd_accept ret val fix


fdesbiens:

fdesbiens:
Hi @mzgrebnak.

Thanks for the PR. I checked this locally on `b6cb589`.

The focused `v4_full_build` builds successfully, and `git diff --check origin/master...HEAD` is clean. However, I do not think this is ready to merge yet because it introduces BSD socket regressions.

**Main issues:**

1. `accept()` now returns the error from `nx_bsd_tcp_create_listen_socket()` after a connection has already been accepted. If recreating the next listen socket fails because BSD socket slots are exhausted, the already-accepted secondary socket should not be dropped. This currently breaks the  socket-exhaustion accept tests:
     - `netx_bsd_tcp_accept_nonblocking_test`
     - `netx_bsd_tcp_accept_nonblocking_timeout_test`

2. The `select()` `writefds` scan no longer reports sockets that are not in use. Existing BSD tests expect closed descriptors to be reflected in all
  requested fd sets. This causes failures in:
     - `netx_bsd_tcp_basic_blocking_test`
     - `netx_bsd_tcp_rcvbuf_test`
     - `netx_bsd_getaddrinfo_test`

3. The new urgent-data handling needs a BSD regression test. The current implementation checks `nx_tcp_socket_receive_queue_head` during the `exceptfds` scan, but `select()` scans `readfds` first and may remove/cache the TCP packet before `exceptfds` is evaluated. That can miss exception readiness when both read and exception sets are supplied.

Please address these regressions and add coverage for the urgent-data/`exceptfds` behavior.

fdesbiens (changes requested):
See my comment for feedback.

PR #376, opened 2026-04-09
BSD recvfromto support

mzgrebnak:
- recvfromto support
- nx_bsd_inet_aton, disabling other format than four numbers separated by dots. Otherwise used in nx_bsd_inet_pton It accepts unallowed format of IP address.


fdesbiens:

fdesbiens (changes requested):
Please make changes per my comment.

fdesbiens:
Hi @mzgrebnak.

Thanks for the PR. I checked this locally on `4aa8e9a`.

This is not ready to merge yet because `./test/cmake/netxduo/run.sh build v4_full_build` fails with `addons/BSD/nxd_bsd.h:241:33: error: no previous declaration for 'recvfromto' [-Werror=missing-declarations]`. 

`nx_bsd_recvfromto()` is defined in `addons/BSD/nxd_bsd.c`, and the native-name macro maps it to `recvfromto`, but `addons/BSD/nxd_bsd.h` does not declare the API near the other BSD receive prototypes.

I also noticed two follow-up issues to address after the build is fixed:
- The `inet_aton()` change rejects abbreviated IPv4 forms that existing tests currently require, such as `11.10.2570` and `11.657930` in `netx_bsd_aton_test.c`. If this behaviour change is intentional, the PR should explain it and update the tests; otherwise, this hunk should be reverted.
- `recvfromto()` only populates `toAddr` for `AF_INET`. For IPv6 UDP sockets, it appears to return success while leaving the destination address unchanged.

Please add focused regression coverage for `recvfromto()` as well, including destination address/port validation.

PR #392, opened 2026-06-16
Additional BSD & select improvements

mzgrebnak:
**nx_bsd_sendto**
Aligning with POSIX. Const parameter before struct nx_bsd_sockaddr *destAddr.

**nx_bsd_soc_close**
Unlisten in case of failure in accepting connection due to resource limitation.

**nx_bsd_inet_aton_impl**
Aligning with POSIX. Use strict parsing to allow notation only with 4 numbers and 3 dots for ipv4.

**nx_bsd_select**
Aligning with POSIX. Optimizing writefds.
Fixing exceptfds.

**nx_bsd_select_wakeup**
Fixing exceptfds.

**nx_bsd_inet_pton**
Aligning with POSIX. Use strict parsing to allow notation only with 4 numbers and 3 dots for ipv4.


addons/BSD/nxd_bsd.c
```diff
@@ -3687,7 +3687,7 @@ NX_BSD_SOCKET *bsd_socket_ptr;
 /*    Application Code                                                    */
 /*                                                                        */
 /**************************************************************************/
-INT  nx_bsd_sendto(INT sockID, CHAR *msg, INT msgLength, INT flags,  struct nx_bsd_sockaddr *destAddr, INT destAddrLen)
+INT  nx_bsd_sendto(INT sockID, CHAR *msg, INT msgLength, INT flags, const struct nx_bsd_sockaddr *destAddr, INT destAddrLen)
```
fdesbiens, line 3724: > Adding `const` here is correct and matches POSIX, but it isn't propagated to the two internal
> helpers that `sendto` forwards `destAddr` to, so this breaks the build.
>
> Both helpers still take a non-const pointer:
> - declarations at `nxd_bsd.c:178` and `nxd_bsd.c:184`
> - definitions at `nxd_bsd.c:12760` and `nxd_bsd.c:13064`
> - call sites at `nxd_bsd.c:3762` (`NX_BSD_RAW_PPPOE_SUPPORT`) and `nxd_bsd.c:3765`
>   (`NX_BSD_RAW_SUPPORT`)
>
> Compiling `nxd_bsd.c` with the CI flags and `-DNX_BSD_RAW_SUPPORT`:
>
> ```
> PR head : nxd_bsd.c:3765:101: warning: passing argument 5 of
>           ‘_nx_bsd_hardware_internal_sendto’ discards ‘const’ qualifier from
>           pointer target type [-Wdiscarded-qualifiers]
> dev     : clean
> ```
>
> The library target is built with `PRIVATE -Werror -Wall -Wextra`
> (`test/cmake/netxduo/CMakeLists.txt:393`) and `v6_bsd_raw_build` defines `NX_BSD_RAW_SUPPORT`
> (`test/cmake/netxduo/CMakeLists.txt:138` and `:327`), so this is a hard failure in that
> configuration rather than just a warning.
>
> Please add `const` to the two helper declarations and definitions.
>
> While you're there: the casts inside this function now discard `const` too —
> `nxd_bsd.c:3838`, `:3839` and `:3852`–`:3857` cast a `const struct nx_bsd_sockaddr *` to
> `struct nx_bsd_sockaddr_in *` / `..._in6 *`. That is MISRA C 2012 **Rule 11.8** (required: a cast
> shall not remove a `const` qualification). They should become
> `(const struct nx_bsd_sockaddr_in *)` etc.

addons/BSD/nxd_bsd.c
```diff
@@ -6002,6 +6009,11 @@ UINT  dot_flag;
            value |= ip_address_number[i] << (24 - (i*8));
        }
     }
+    else if (use_strict_parsing == NX_TRUE)
```
fdesbiens: > The strict/lenient split itself is a good idea, but it's being applied to the wrong function.
>
> `inet_aton()` is specified by 4.3BSD and implemented by glibc, musl and the BSDs to accept all
> four forms — `a`, `a.b`, `a.b.c`, `a.b.c.d`. The `inet_aton(3)` man page documents this
> explicitly. `inet_pton()` is the function that accepts only the full dotted-quad form. So making
> `inet_aton` strict moves *away* from POSIX/BSD behaviour rather than toward it, and it is a
> silent behaviour change for every existing application that relies on the documented forms.
>
> It also makes the two entry points inconsistent with each other: `nx_bsd_inet_addr` is routed
> with `NX_FALSE` at `nxd_bsd.c:6103`, so after this change
> `inet_addr("11.10.2570")` succeeds while `inet_aton("11.10.2570")` fails. On every real platform
> `inet_addr()` is a thin wrapper over `inet_aton()` and they share one parser, so this divergence
> would be surprising.
>
> Suggested change: pass `NX_FALSE` from `inet_aton` (keeping today's behaviour) and `NX_TRUE`
> only from the two `inet_pton` call sites at `nxd_bsd.c:13557` and `nxd_bsd.c:13666`. That also
> lets the three test changes in this PR be reverted (see the comments on
> `netx_bsd_aton_test.c:65` and `netx_bsd_recvfromto_test.c:247`).
>
> Minor: `return 0;` should be `return(0);` to match every other return in this function, and
> `use_strict_parsing == NX_TRUE` is better written as a plain truth test since the parameter is an
> `INT` supplied by internal callers.

addons/BSD/nxd_bsd.h
```diff
@@ -1004,7 +1004,7 @@ INT  nx_bsd_getsockname(INT sockID, struct nx_bsd_sockaddr *localAddress, INT *a
 INT  nx_bsd_ioctl(INT sockID, INT command, INT *result);
 nx_bsd_in_addr_t nx_bsd_inet_addr(const CHAR *buffer);
 CHAR *nx_bsd_inet_ntoa(struct nx_bsd_in_addr address_to_convert);
-INT  nx_bsd_inet_aton(const CHAR *cp_arg, struct nx_bsd_in_addr *addr);
+INT  nx_bsd_inet_aton_impl(const CHAR *address_buffer_ptr, struct nx_bsd_in_addr *addr, INT use_strict_parsing);
```
fdesbiens: > This renames a public API function without leaving anything behind under the old name. After this
> change no `nx_bsd_inet_aton` symbol exists anywhere — `git grep nx_bsd_inet_aton` on the PR head
> returns only the `_impl` spelling.
>
> That breaks applications built with `NX_BSD_ENABLE_NATIVE_API`. That option is documented and
> user-selectable (`nxd_bsd.h:117`), and when it is defined the override block at `nxd_bsd.h:180` is
> skipped entirely — so those applications call `nx_bsd_inet_aton()` directly, and there is no
> replacement for them at all, strict or lenient. The new `inet_aton` macro doesn't help, because
> it lives inside the block that native mode compiles out.
>
> Suggestion that avoids all of this: make the strict/lenient worker a `static` helper (e.g.
> `nx_bsd_inet_aton_internal(const CHAR *, struct nx_bsd_in_addr *, UINT strict)`) and keep
> `nx_bsd_inet_aton()` as the public function, unchanged in signature and behaviour, calling the
> helper with the lenient flag. `nx_bsd_inet_pton()` then calls the helper with the strict flag.
> The header needs no change at all, `NX_BSD_ENABLE_NATIVE_API` keeps working, and the internal
> helper stays out of the public surface.

addons/BSD/nxd_bsd.h
```diff
@@ -233,7 +233,7 @@ extern   "C" {
 #define nx_bsd_ioctl            ioctl
 #define nx_bsd_inet_addr        inet_addr
 #define nx_bsd_inet_ntoa        inet_ntoa
-#define nx_bsd_inet_aton        inet_aton
+#define inet_aton(address_buffer_ptr, addr) nx_bsd_inet_aton_impl(address_buffer_ptr, addr, NX_TRUE)
```
fdesbiens: > Three problems with this line, all of which disappear if the `static` helper approach in the
> `nxd_bsd.h:1007` comment is used instead.
>
> 1. **It reverses the direction of the block.** Every other entry in this section maps the
>    internal `nx_bsd_*` name *onto* the plain POSIX name (`#define nx_bsd_inet_addr inet_addr`),
>    which is how the `nx_bsd_`-prefixed definitions in `nxd_bsd.c` come to be compiled under their
>    POSIX names. This line goes the other way, defining `inet_aton` in terms of an
>    `nx_bsd_*` symbol, so the file now contains two opposite conventions in adjacent lines.
>
> 2. **It is the only function-like macro in a block of object-like macros.** As a result
>    `inet_aton` is no longer usable as a value — `&inet_aton`, or passing it as a callback, stops
>    compiling for application code that previously worked.
>
> 3. **The macro parameters are not parenthesised.** MISRA C 2012 **Rule 20.7** (required) requires
>    macro parameters to be wrapped in parentheses in the expansion, i.e.
>    `nx_bsd_inet_aton_impl((address_buffer_ptr), (addr), NX_TRUE)`.

addons/BSD/nxd_bsd.c
```diff
@@ -13484,7 +13554,7 @@ struct  nx_bsd_in_addr ipv4_addr;
     if(af == AF_INET)
     {
         /* Convert IPv4 address from presentation to numeric. */
-        if(nx_bsd_inet_aton(src, &ipv4_addr))
+        if(nx_bsd_inet_aton_impl(src, &ipv4_addr, NX_TRUE))
```
fdesbiens: > `use_strict_parsing` only enforces the number of dot-separated groups. It doesn't restrict the
> digit format, so `inet_pton(AF_INET, ...)` still accepts input that POSIX requires it to reject:
>
> - hex groups — `"0xb.0xa.0xa.0xa"`
> - octal groups and leading zeros — `"013.012.012.012"`, `"127.0.0.010"`
> - trailing whitespace, accepted by the `nx_bsd_isspace` check at `nxd_bsd.c:5966`
>
> The existing assertions at `netx_bsd_aton_test.c:69-83` confirm the hex and octal forms still
> return 1, and those go through the same parser that `inet_pton` now uses with `NX_TRUE`.
>
> For strict mode to actually mean "POSIX `inet_pton`", it also needs to require decimal digits
> only, reject a leading `0` on a multi-digit group, and reject trailing whitespace. Worth adding
> test cases for each of those, since they're the interesting inputs.

addons/BSD/nxd_bsd.c
```diff
@@ -5008,6 +5008,13 @@ UINT                 index;
 
         }
 
+        if(bsd_socket_ptr -> nx_bsd_socket_status_flags & NX_BSD_SOCKET_SERVER_MASTER_SOCKET
+            && (bsd_socket_ptr -> nx_bsd_socket_union_id).nx_bsd_socket_secondary_socket_id == NX_BSD_MAX_SOCKETS)
+        {
+            /* Unlisten in case of failure in accepting connection due to resource limitation. */
+            nx_tcp_server_socket_unlisten(nx_bsd_default_ip, tcp_socket_ptr -> nx_tcp_socket_port);
+        }
```
fdesbiens: > The underlying bug is real and worth fixing: when `accept()` cannot allocate a replacement
> secondary socket it sets the master's id to `NX_BSD_MAX_SOCKETS` (`nxd_bsd.c:3020-3026`), which
> makes the cleanup block at `nxd_bsd.c:5027` skip itself, and the NetX listen structure for the
> port is then never released. Good catch.
>
> The guard needs to be stronger though. The pre-existing path deliberately scans for other master
> sockets sharing the same secondary socket before it unlistens (`nxd_bsd.c:5030-5046`, unlisten at
> `:5096`). This new call has no such check, so if two master sockets are bound to the same port
> with `SO_REUSEADDR` and share a secondary socket, closing the one whose `accept()` failed will
> tear down the listen that the *other* master socket is still using — its subsequent `accept()`
> then has no listen structure to work with. Please reuse the same "am I the last master on this
> port" scan, or at minimum verify no other in-use master socket has `NX_BSD_SOCKET_ENABLE_LISTEN`
> on `nx_bsd_socket_local_port`.
>
> Two smaller points on the same condition:
>
> - `== NX_BSD_MAX_SOCKETS` should be `>= NX_BSD_MAX_SOCKETS` to match how the rest of the file
>   tests for an invalid secondary id (`nxd_bsd.c:5027`, `nxd_bsd.c:8189`).
> - `a & MASK && (b)` mixes `&` and `&&` without parentheses. It parses as intended, but it is
>   MISRA C 2012 **Rule 12.1** (advisory) and every comparable condition in this file
>   parenthesises each operand. Please write it as
>   `if ((flags & NX_BSD_SOCKET_SERVER_MASTER_SOCKET) && (id >= NX_BSD_MAX_SOCKETS))`, with the
>   continuation lined up the way the surrounding conditions are.

addons/BSD/nxd_bsd.c
```diff
@@ -8358,6 +8377,27 @@ INT                     ret;
 
                 NX_BSD_FD_SET(i + NX_BSD_SOCKFD_START, &exceptfds_found);
             }
+
```
fdesbiens, line 8502: > This reads `nx_tcp_socket_receive_queue_head` and then dereferences the packet's
> `nx_packet_prepend_ptr` while holding only the BSD protection mutex.
>
> The TCP receive queue is manipulated under `TX_DISABLE` — both by the IP thread in `_nx_tcp_socket_packet_process` and by `_nx_tcp_socket_receive`. So another application thread calling `recv()` on this same socket can dequeue the packet and release it back to the pool between the read on this line and the dereference two lines later, which makes the `tcp_header_ptr` access a use-after-free.
>
> Wrapping the read and the header check in `TX_DISABLE` / `TX_RESTORE` fixes it and matches what the rest of the stack does. Note the near-identical block you added in `nx_bsd_select_wakeup` is already inside a `TX_DISABLE` region, which is why only this copy is exposed.
>
> Also, for consistency with the file's style, please declare `tcp_header_ptr` the way the
> equivalent block at `nxd_bsd.c:8236` does, and use ` -> ` rather than `->`.
>
> One limitation worth a comment in the code: this scan only sees a URG packet while it is still on the TCP receive queue. Once a previous `select()` readfds pass has moved it into `nx_bsd_socket_received_packet` (`nxd_bsd.c:8221`) the queue no longer holds it, so pending out-of-band data becomes invisible to `exceptfds`. Not necessarily something to fix here, but better documented than discovered later.

addons/BSD/nxd_bsd.c
```diff
@@ -8298,13 +8310,17 @@ INT                     ret;
                 NX_BSD_FD_SET(i + NX_BSD_SOCKFD_START, &writefds_found);
             }
 
-            /* Check to see if there is a connection request pending on a client socket.  */
-            else if ((nx_bsd_socket_array[i].nx_bsd_socket_status_flags & NX_BSD_SOCKET_CONNECTION_REQUEST) &&
-                     !(nx_bsd_socket_array[i].nx_bsd_socket_status_flags & NX_BSD_SOCKET_SERVER_MASTER_SOCKET))
+            /* Check to see if there is a connection request pending.  */
+            else if (nx_bsd_socket_array[i].nx_bsd_socket_status_flags & NX_BSD_SOCKET_CONNECTION_REQUEST)
```
fdesbiens: > Excluding a listening socket from `writefds` is the right behaviour, but keying it off
> `NX_BSD_SOCKET_CONNECTION_REQUEST` makes it configuration-dependent.
>
> The master socket only ever gets `CONNECTION_REQUEST` set at `nxd_bsd.c:8693`, which is inside `#ifndef NX_DISABLE_EXTENDED_NOTIFY_SUPPORT`. With extended notify disabled, a master socket with a pending connection is still marked `NX_BSD_SOCKET_CONNECTED` (`nxd_bsd.c:8692`) but has no `CONNECTION_REQUEST`, so it falls through to the `else if (tcp_socket)` branch you added at `nxd_bsd.c:8333` and is reported writable again. Test 5 in this PR would not catch that, because it only runs in the default configuration.
>
> Testing `NX_BSD_SOCKET_SERVER_MASTER_SOCKET` directly — either as its own early branch, or additionally inside the TCP branch at `:8333` — is both simpler and independent of the notify configuration.
>
> Separately, hoisting the `CONNECTION_REQUEST` test above the `NX_BSD_SOCKET_ERROR` test changes behaviour for one case: a master socket that has both `CONNECTION_REQUEST` and `ERROR` set used to reach the `ERROR` branch and be reported writable, and now returns nothing. POSIX requires a socket with a pending error to be reported writable, so the `ERROR` check should stay ahead of the others in this chain.

addons/BSD/nxd_bsd.c
```diff
@@ -10290,6 +10331,35 @@ INT                     bsd_sock_id;
...
+                        NX_TCP_HEADER   *tcp_header_ptr;
+
+                            tcp_header_ptr =  (NX_TCP_HEADER *)packet_ptr -> nx_packet_prepend_ptr;
+                            if (tcp_header_ptr -> nx_tcp_header_word_3 & NX_TCP_URG_BIT)
+                            {
+                                NX_BSD_FD_SET(bsd_sock_id, &suspend_info -> nx_bsd_socket_suspend_exception_fd_set);
+                            }
+                        }
+                    }
+                }
```
fdesbiens: > I don't think this block adds behaviour, and it costs interrupt latency in a place where that
> matters.
>
> NetX already raises a dedicated notification for every URG packet: `_nx_tcp_socket_packet_process` invokes the socket's urgent callback `common/src/nx_tcp_socket_packet_process.c:467-484`), which for BSD sockets is `nx_bsd_tcp_exception_notify` (registered at `nxd_bsd.c:940`, defined at `:8563`). That calls `nx_bsd_select_wakeup(index, FDSET_EXCEPTION)` at `nxd_bsd.c:8587`, which sets exactly the same `nx_bsd_socket_suspend_exception_fd_set` bit this new code sets, for the same packet.
>
> And a thread that calls `select()` when a URG packet is already queued never suspends in the first place — the new exceptfds scan at `nxd_bsd.c:8384` satisfies it immediately.
>
> Meanwhile this code runs inside the `TX_DISABLE` region of the wakeup loop
> (`nxd_bsd.c:10266`–`:10323`), so it adds pointer chasing to a critical section with interrupts
> off, on every read notification for a socket that appears in both `readfds` and `exceptfds`. Given our emphasis on runtime speed and code size, I'd drop the block unless you have a concrete case where `nx_bsd_tcp_exception_notify` doesn't fire — in which case that case is worth a regression test.
>
> The addition of `nx_bsd_tcp_exception_notify` to the `CALLED BY` list at `nxd_bsd.c:10285` is
> correct and worth keeping either way.

addons/BSD/nxd_bsd.c
```diff
@@ -5812,7 +5819,7 @@ UINT    size;
 /*    Application Code                                                    */
 /*                                                                        */
 /**************************************************************************/
-INT nx_bsd_inet_aton(const CHAR *address_buffer_ptr, struct nx_bsd_in_addr *addr)
+INT nx_bsd_inet_aton_impl(const CHAR *address_buffer_ptr, struct nx_bsd_in_addr *addr, INT use_strict_parsing)
```
fdesbiens: > The comment block above this function (`nxd_bsd.c:5780-5814`) still describes the old function: its `FUNCTION` line reads `inet_aton`, and the `INPUT` section lists only `address_buffer_ptr` and `addr`, so the new `use_strict_parsing` parameter is undocumented.
>
> We require new functions and parameters to be documented in the same style as the surrounding code, so please add `use_strict_parsing` to `INPUT` with a note on what each value means, and update the `FUNCTION` name and the `DESCRIPTION` to mention the strict mode. (If the `static` helper approach from the `nxd_bsd.h:1007` comment is adopted, the new helper needs its own block in this format.)

test/regression/bsd_test/netx_bsd_aton_test.c
```diff
@@ -59,12 +59,15 @@ int    status;
     if((status !=  1) || (in_val.s_addr != htonl(0x0b0a0a0a)))
         error_counter++;
 
+    /* Abbreviated forms (fewer than four dotted numbers) are now rejected by
+       inet_aton(), which performs strict parsing. The a.b.c form must fail. */
     status = inet_aton("11.10.2570", &in_val); /* 2570 == 0x0a0a */
-    if((status !=  1) || (in_val.s_addr != htonl(0x0b0a0a0a)))
+    if(status != 0)
```
fdesbiens: > These two assertions were passing before this PR, and they were testing documented `inet_aton()`
> behaviour — the `a.b.c` and `a.b` forms are part of the function's contract on every platform
> that has it. Rewriting them to expect failure encodes the behaviour change rather than
> testing it.
>
> If `inet_aton` stays lenient as suggested on `nxd_bsd.c:6012`, this hunk can simply be reverted.
> The strict cases belong in the `inet_pton` tests instead, where rejecting `"11.10.2570"` is
> correct — and ideally alongside the hex/octal/leading-zero cases from the `nxd_bsd.c:13557`
> comment

test/regression/bsd_test/netx_bsd_recvfromto_test.c
```diff
@@ -237,19 +237,19 @@ UINT                        actual_status;
...
+    /* --- Test 3: inet_aton() now rejects abbreviated IPv4 forms (strict parsing) --- */
     {
     struct in_addr  in_val;
     INT             rc;
 
-        /* a.b format: 11.657930 == 11.0x0a0a0a == 11.10.10.10 */
+        /* a.b format (11.657930) must be rejected under strict parsing. */
         rc = inet_aton("11.657930", &in_val);
-        if (rc == 0)
+        if (rc != 0)
```
fdesbiens: > Same point as on `netx_bsd_aton_test.c:65`: this block was added specifically to assert that
> `inet_aton()` still accepts the abbreviated forms, and it's now inverted to assert the opposite.
> Please revert this hunk along with the `inet_aton` behaviour.

test/regression/bsd_test/netx_bsd_select_improvements_test.c
```diff
@@ -23,6 +23,14 @@
...
  *  3. A socket that is not in use must be reported in writefds (regression for
  *     the writefds "not in use" restore).
+ *
+ *  4. A TCP socket that is bound but NOT connected must NOT be reported in
+ *     writefds (a non-connected TCP socket is not writable regardless of its
+ *     bind state).
+ *
+ *  5. A listening (master) socket with a pending connection must be reported
+ *     readable (ready to accept) but MUST NOT be reported writable, even though
+ *     the master is internally marked CONNECTED once a client connects.
```
fdesbiens: > Tests 4 and 5 cover the two `writefds` changes nicely, but the other half of this PR — the
> `exceptfds` URG scan in `select()` (`nxd_bsd.c:8380-8400`) and the URG check in
> `nx_bsd_select_wakeup` (`nxd_bsd.c:10334-10361`) — has no test at all. We need
> matching regression tests and 100% coverage for new or extended behaviour.
>
> Worth adding at least: a socket in `exceptfds` only (no `readfds`) that becomes ready when an
> urgent segment arrives, and the negative case where normal data must not set `exceptfds` — the
> latter partly exists as Test 2, but only for the `readfds`+`exceptfds` combination.
>
> Note that the strict-parsing branch at `nxd_bsd.c:6012` *is* covered, and the lenient `a.b` /
> `a.b.c` branches keep their coverage through `netx_bsd_inet_addr_pton_test.c:73-77`, so no
> coverage is lost there.
>
> Also worth confirming: does the `unlisten` fix at `nxd_bsd.c:5011` have a test? The failed-accept
> path it depends on is hard to reach, but without one the fix is untested.

test/regression/bsd_test/netx_bsd_pton_test.c
```diff
@@ -120,8 +120,9 @@ UCHAR ipv4_addr[4]; /* 4*8 == 32 */
     if(status != 0)
         error_counter++;
 
+    /* inet_pton shall not accept abbreviated IPv4 addresses */
     status = inet_pton(AF_INET,  ipv4_addr_str3, ipv4_addr);
-    if((status != 1) || (memcmp(ipv4_addr, ipv4_addr_num3, 4) != 0 ))
+    if((status == 1) || (memcmp(ipv4_addr, ipv4_addr_num3, 4) == 0 ))
```
fdesbiens: > The `memcmp` half of this condition doesn't test anything meaningful. When `inet_pton` rejects an
> address it returns 0 without writing to `dst`, so `ipv4_addr` still holds the result of the last
> successful call — `ipv4_addr_str1` = `"1.2.3.4"` from line 111. The assertion therefore passes
> only because `0x01020304` happens to differ from `ipv4_addr_num3` (`0x0b0a0a0a`); reorder the
> fixtures and it would start failing for no real reason.
>
> `if (status != 0) error_counter++;` is the whole check.
>
> (Unrelated pre-existing bug, since you're in the area: `netx_bsd_inet_addr_pton_test.c:111` uses
> `&&` where `||` was clearly intended, so that assertion can never fail. Fine to leave for a
> separate PR.)

test/regression/bsd_test/netx_bsd_select_improvements_test.c
```diff
@@ -162,6 +170,43 @@ char                buf[64];
     /* Signal client thread to connect. */
     tx_semaphore_put(&sema_1);
 
+    /* ================================================================
```
fdesbiens, line 177: > Test 5 is defined here and Test 4 about 130 lines further down at line 302, so the file now reads
> 1, 2, 3, 5, 4. The ordering is forced by where each test has to run relative to `accept()`, which
> is fine — just swap the two numbers so they read in order.

fdesbiens (changes requested):
Thanks for this — there are four independent improvements bundled here, and three of them are genuinely good catches. Excluding a listening socket from `writefds`, excluding a bound-but-unconnected TCP socket from `writefds`, and the `unlisten` on the failed-accept path are all real POSIX/resource bugs, and the added regression tests are welcome.

Two things need to change before this can land, and I'd like to discuss a third:

1. **It breaks the `v6_bsd_raw_build` configuration.** The `const` added to `nx_bsd_sendto` is not  propagated to the two internal helpers it forwards to, and the library is compiled with `-Werror`. See the comment on `nxd_bsd.c:3690` — this is a one-line-per-site fix.

2. **Making `inet_aton()` strict is the wrong direction.** `inet_aton()` is specified by 4.3BSD and implemented by glibc/musl/BSD to *accept* `a`, `a.b`, `a.b.c` and `a.b.c.d`; `inet_pton()` is the strict one. The PR makes `inet_aton` reject the abbreviated forms while leaving `inet_addr` lenient, which is also self-inconsistent, and it rewrites three previously-passing tests to expect the new failures. Strictness should be applied only at the `inet_pton` call sites. Details on `nxd_bsd.c:6012`.

Related: the rename to `nx_bsd_inet_aton_impl` removes `nx_bsd_inet_aton` from the public API, which breaks applications built with `NX_BSD_ENABLE_NATIVE_API` (`nxd_bsd.h:1007`), and the new function-like macro inverts the direction of the entire override block (`nxd_bsd.h:236`).

3. **The `nx_bsd_select_wakeup` change looks redundant** — NetX already delivers a URG notification through `nx_bsd_tcp_exception_notify`. See `nxd_bsd.c:10334`; happy to be corrected if you have a case in mind that the existing callback misses.

There are also a couple of correctness issues in the new `select()` exceptfds scan
(`nxd_bsd.c:8388`) and in the `unlisten` guard (`nxd_bsd.c:5011`), plus one test assertion that passes by coincidence (`netx_bsd_pton_test.c:125`).

Two project-policy items: the new `use_strict_parsing` parameter needs to be documented in the function's comment block , and the `nx_bsd_sendto` signature change needs a matching documentation update in [rtos-docs-asciidoc](https://github.com/eclipse-threadx/rtos-docs-asciidoc).

Finally, a general style note: the new code uses `if(` and `a->b`, whereas this file consistently uses `if (` and `a -> b`. Please match the surrounding style.

mzgrebnak:
I tried to address all your comments. The main changes are:

- **OOB / select EXC part** — the whole solution is different now. There is a new mechanism for receiving and selecting OOB data.
- **Brand-new socket limit tests and OOB tests** (AI-generated):
netx_bsd_tcp_accept_socket_limit_shared_test
netx_bsd_tcp_accept_socket_limit_test
netx_bsd_tcp_oob_errno_test
netx_bsd_tcp_oob_test
- **New socket option** SO_TYPE.

addons/BSD/nxd_bsd.c
mzgrebnak: I believe it covered the case when the socket was in both the read and exception fd sets and select() was blocking. I agree the latency cost was significant. In any case, right now the OOB mechanism is entirely different.

fdesbiens:
Hi @mzgrebnak.

Thanks for the latest updates. I completed an additional maintainer review and pushed commit a260d626 directly to this
PR branch.

The follow-up work:

- Made TCP urgent-byte extraction safe across chained NX_PACKETs.
- Corrected urgent-pointer handling when TCP trims overlapping data.
- Installed OOB state before waking a higher-priority thread blocked in select().
- Implemented non-consuming MSG_OOB | MSG_PEEK behavior.
- Ensured an urgent pointer extending into a later segment is not prematurely recorded as consumed.
- Added validation for NX_BSD_SOCKET_OOB_QUEUE_SIZE.
- Expanded regression coverage for packet chains, overlap trimming, select() wake ordering, future-segment urgent data,
  and MSG_PEEK.
- Aligned the affected headers and added code with the project style and attribution requirements.

I also audited the earlier review threads; the previous const-correctness, inet_aton/inet_pton, native API, shared-
listener cleanup, UAF, writefds, and regression-test comments remain addressed.

Local validation used GCC 14 with CMake and Ninja. All 10 PR-affected regressions passed, followed by the complete BSD
set: 53 passed and 2 were configuration-based N/A. The touched objects also compile with packet chaining disabled.

The missing documentation was prepared separately in the canonical documentation repository. CI for this PR is now
running.

I look forward to your next contributions! I will merge to dev once CI is green; this will ship with the 6.5.2 release in September.

Issue #161, opened 2023-03-23
Supporting RSA signed client certificates with TLS 1.3


jkuwaha:
**Is your feature request related to a problem? Please describe.**
With TLS 1.3, TLS session cannot be established with clients using RSA signed certificates.

**Describe the solution you'd like**
It would be good if NetX supports RSA signed client certificates with TLS 1.3.

TiejunMS:
@jkuwaha , thanks for opening this feature request. This feature is in our backlog. It is not prioritized at the moment.

jkuwaha:
@TiejunMS,
Thank you for the information.
We look forward to being supported (hopefully within a year).

GKNSTM:
Hello. Any news on this feature request ? we would be very interested to have this RSA certificate support in TLS 1.3 in Netxduo.

jespertragardh:
Mostly for making this more visible in search engines.

To my understanding, this applies to both client and server certificates.

The technical problem is that for RSA, NetX Duo supports only the RSA signature algorithms (as can be seen in the _ClientHello_ message for a TLS 1.3 connection):

* rsa_pkcs1_sha256 (0x0401)
* rsa_pkcs1_sha384 (0x0501)
* rsa_pkcs1_sha512 (0x0601)

But these are _RSASSA-PKCS1-v1_5_ algorithms and according to RFC 8446 sect. 4.4.3 these may not be used for the _CertificateVerify_ protocol message (send by server or client) regardless of how the certificate chain is signed. _CertificateVerify_ must use _RSASSA-PSS_ algorithms, one of:

* rsa_pss_pss_sha256(0x0809),
* rsa_pss_pss_sha384(0x080a),
* rsa_pss_pss_sha512(0x080b),

and until they are implemented, you cannot use RSA certificates with TLS 1.3 in NetX Duo.
You can you use ECC certificates so that _CertificateVerify_ can be signed with ecdsa_secp256r1_sha256 and such a certificate can be signed with a RSA-based certificate (e.g. from the CA or its intermediates) because validation of the certificate chain is a different matter from verifying the signature of the _CertificateVerify_ protocol message.

Formally, NetX Duo is not compliant with TLS 1.3 since RFC 8446 sect. 9.1 says:

> [9.1](https://www.rfc-editor.org/info/rfc8446/#section-9.1).  Mandatory-to-Implement Cipher Suites
>
>   In the absence of an application profile standard specifying
>   otherwise:
>
[...]

PR #399, opened 2026-07-08
Added RSA-PSS signing support for TLS 1.3 client mutual authentication with an RSA certificate

EdouardMALOT:
## Summary

This PR completes the RSA-PSS support introduced in #377 by adding the **signing** side: a TLS 1.3 client can now perform mutual authentication (client certificate) with an RSA certificate.

#377 added RSA-PSS **verification**, which lets a NetX Duo client verify the CertificateVerify sent by an RSA server. However, when the server requests a client certificate, the client must also **sign** its own CertificateVerify with RSA-PSS — [RFC 8446 §4.4.3](https://www.rfc-editor.org/rfc/rfc8446#section-4.4.3) forbids RSASSA-PKCS1-v1_5 in CertificateVerify. Before this PR, `_nx_secure_tls_process_certificate_request()` returned `NX_SECURE_TLS_UNSUPPORTED_CERT_SIGN_TYPE` for any RSA local certificate in TLS 1.3, making client-cert authentication impossible with RSA keys.

This implements `rsa_pss_rsae_sha256` (0x0804), which is the mandatory-to-implement signature algorithm for CertificateVerify per [RFC 8446 §9.1](https://www.rfc-editor.org/rfc/rfc8446#section-9.1). Related: #161.

## Changes

- **`crypto_libraries/inc/nx_crypto_rsa.h`** — declare `_nx_crypto_rsa_pss_sign()`
- **`crypto_libraries/src/nx_crypto_rsa.c`** — implement `_nx_crypto_rsa_pss_sign()`: EMSA-PSS-ENCODE ([RFC 8017 §9.1.1](https://www.rfc-editor.org/rfc/rfc8017#section-9.1.1)), salt length == hash length as required by [RFC 8446 §4.2.3](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.3), salt generated via `NX_CRYPTO_RAND()` (same entropy source as the ECDSA nonce path). The RSA private-key operation is then applied to the encoded message by the existing signing flow.
- **`nx_secure/src/nx_secure_tls_process_certificate_request.c`** — TLS 1.3: instead of rejecting RSA local certificates, select `rsa_pss_rsae_sha256` and proceed with the client Certificate/CertificateVerify flow
- **`nx_secure/src/nx_secure_tls_send_certificate_verify.c`** — TLS 1.3: build the CertificateVerify message with the SignatureScheme wire code (`rsa_pss_rsae_sha256/384/512` handled) and the PSS-encoded message, instead of PKCS#1 v1.5 padding (which remains used for TLS 1.2)

## Notes / limitations

- When answering a CertificateRequest, the client currently selects `rsa_pss_rsae_sha256` unconditionally (the RFC 8446 §9.1 MUST algorithm, supported by any compliant server). The CertificateVerify encoding path itself also supports `rsa_pss_rsae_sha384/512` should the selection be extended later.
- `rsa_pss_pss_*` variants (RSASSA-PSS certificates, `0x0809`–`0x080b`) are out of scope; this covers the common case of classic RSA certificates (`rsae`).

## Test plan

- [x] TLS 1.3 mutual authentication handshake (server requests a client certificate) completes with an RSA client certificate, `rsa_pss_rsae_sha256` CertificateVerify accepted by the server (tested against a Mosquitto/OpenSSL broker requiring client certificates)
- [x] TLS 1.3 server-only handshake (no CertificateRequest) unaffected
- [x] TLS 1.2 client certificate path still uses RSASSA-PKCS1-v1_5 (unchanged behavior)


EdouardMALOT:
@fdesbiens : the remaining Secure_Interoperability / run_tests failure is unrelated to this change — the job fails during the "Install softwares" step.

fdesbiens:
Thank you for this PR, @EdouardMALOT, and for all the others you submitted. 

I am currently on vacation; I will review them once I am back, the week of July 20.

fdesbiens:
> @fdesbiens : the remaining Secure_Interoperability / run_tests failure is unrelated to this change — the job fails during the "Install softwares" step.

This is expected. The script relies on ifconfig to tweak networking configuration, and this should be modernised. I will spend some time fixing this along with the reviews.

crypto_libraries/src/nx_crypto_rsa.c
```diff
@@ -752,3 +752,208 @@ static const UCHAR _pss_zero8[8] = {0, 0, 0, 0, 0, 0, 0, 0};
...
+/*    scratch_length      Size of scratch in bytes                       */
+/*                                                                        */
+/*  OUTPUT                                                                */
+/*                                                                        */
+/*    NX_CRYPTO_SUCCESS                EM built successfully             */
+/*    NX_CRYPTO_NOT_SUCCESSFUL         em_len smaller than minimum       */
+/*    NX_CRYPTO_INVALID_BUFFER_SIZE    Scratch too small                 */
+/*                                                                        */
+/**************************************************************************/
+UINT _nx_crypto_rsa_pss_sign(const UCHAR *message_hash, UINT hash_length,
```
fdesbiens: > This function writes `em_len = ceil(em_bits/8)` bytes into `em`, but its parameter list has no output-buffer length. Compare `scratch`, which correctly gets a `scratch_length` and is checked at `:823`. There is no equivalent check for `em`, so the caller's buffer size is simply assumed.
>
> The one caller does not satisfy that assumption in all configurations. In `nx_secure_tls_send_certificate_verify.c`, `em` is the 600-byte `_nx_secure_padded_signature` and `em_bits` is `(data_size << 3) - 1`, where `data_size` is `nx_secure_rsa_public_modulus_length` taken straight from the local certificate (`:499`). Nothing in nx_secure bounds that value — I grepped for a cap and there is none. Any RSA client certificate with a modulus above 600 bytes, i.e. above 4800 bits, overruns the buffer.
>
> Reproduced with a 768-byte modulus (6144-bit key) and `em` allocated at exactly 600 bytes:
>
> ```
> em buffer = 600 bytes, modulus = 768 bytes -> writing 768 bytes
> ==56476==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf4e03f5f
> WRITE of size 1 at 0xf4e03f5f thread T0
>     #0 _nx_crypto_sha256_digest_calculate  nx_crypto_sha2.c:317
>     #1 _nx_crypto_method_sha256_operation  nx_crypto_sha2.c:786
>     #2 _nx_crypto_rsa_pss_sign             nx_crypto_rsa.c:908
> 0xf4e03f5f is located 135 bytes after 600-byte region [0xf4e03c80,0xf4e03ed8)
> ```
>
> The first write to land outside is H, written through `h = em + db_len` at `:908`, and the masked-DB loop then writes across the whole range.
>
> This is not attacker-controlled — the modulus comes from the device's own certificate — so it is a robustness and configuration-safety problem rather than a remote issue. It is still memory corruption in a signing path, and a new public crypto API that cannot be called safely without out-of-band knowledge of its own output size.
>
> Two changes would close it, and I would like both:
>
> 1. Add an `em_length` parameter and return `NX_CRYPTO_INVALID_BUFFER_SIZE` when `em_len > em_length`. It is a new function, so there is no compatibility cost to fixing the signature now.
> 2. Guard the call site on `data_size > sizeof(_nx_secure_padded_signature)` and return `NX_SECURE_TLS_INVALID_CERTIFICATE`.
>
> Worth knowing: the PKCS#1 v1.5 path in the same function has the same missing guard already — `working_ptr = &_nx_secure_padded_signature[data_size - signature_length]` at `:649` followed by a 36-byte copy. That one is pre-existing and not yours, but a single `data_size` check placed before the version branch fixes both, which seems like the right shape.

crypto_libraries/src/nx_crypto_rsa.c
```diff
@@ -752,3 +752,208 @@ static const UCHAR _pss_zero8[8] = {0, 0, 0, 0, 0, 0, 0, 0};
...
+    salt = scratch + db_len;
+    h    = em + db_len; /* H sits at em[db_len..em_len-2]. */
+
+    /* Step 4 – generate the random salt. NX_CRYPTO_RAND() is the same
+     * source ECDSA uses for its nonce, so the entropy is at least as good
+     * as the existing signature paths on this device. */
+    for (i = 0u; i < s_len; i++)
+    {
+        salt[i] = (UCHAR)((UINT)NX_CRYPTO_RAND() & 0xFFu);
+    }
```
fdesbiens: > The comment above this loop says `NX_CRYPTO_RAND()` is "the same source ECDSA uses for its nonce, so the entropy is at least as good as the existing signature paths on this device". That is not accurate, and the difference matters.
>
> ECDSA and EC key generation go through `NX_CRYPTO_RBG`, not `NX_CRYPTO_RAND` — see `nx_crypto_ec.c:2945` and `:4671`. `NX_CRYPTO_RBG` resolves to `_nx_crypto_drbg` in `NX_CRYPTO_SELF_TEST` (FIPS) builds and to `_nx_crypto_huge_number_rbg` otherwise (`nx_crypto.h:92` and `:100`). So:
>
> - In a FIPS build, ECDSA gets a real DRBG while this salt gets raw `rand()`. The two are not the same source at all.
> - Even in a default build, `_nx_crypto_huge_number_rbg` consumes all 32 bits of each `NX_CRYPTO_RAND()` result (`nx_crypto_huge_number.c:41-45`), whereas this loop keeps only the low 8 bits and throws away 24. With a classic LCG `rand()` the low bits are the weakest part of the output, so this is the least favourable way to use it.
>
> The fix is shorter than the current code and inherits the right behaviour in every build:
>
> ```c
> status = NX_CRYPTO_RBG(s_len << 3, salt);
> if (status != NX_CRYPTO_SUCCESS)
> {
>     return(status);
> }
> ```
>
> On severity, to be fair to the change: PSS salt does not need to be secret, and unlike an ECDSA nonce a repeated or predictable salt does not leak the private key — it weakens the tightness of the security proof and loses the randomisation benefit, nothing more. So this is a robustness and FIPS-consistency fix, not a key-compromise bug. But there is no reason to bypass the RBG the rest of the library uses, and the comment should not claim parity that does not exist.

crypto_libraries/src/nx_crypto_rsa.c
```diff
@@ -752,3 +752,208 @@ static const UCHAR _pss_zero8[8] = {0, 0, 0, 0, 0, 0, 0, 0};
     return(NX_CRYPTO_SUCCESS);
 }
 
+
+/**************************************************************************/
+/*                                                                        */
+/*  FUNCTION                                               RELEASE        */
+/*                                                                        */
+/*    _nx_crypto_rsa_pss_sign                             PORTABLE C      */
```
fdesbiens, line 794: We need matching regression tests and 100% coverage for new features. There are none here, and gcov confirms nothing reaches the code. From `tls_1_3_enable_build_coverage` after the full 150-test run:
>
> ```
> _nx_crypto_rsa_pss_mgf1      executable= 33  executed=  0  never= 33  (0%)
> _nx_crypto_rsa_pss_verify    executable= 53  executed=  0  never= 53  (0%)
> _nx_crypto_rsa_pss_sign      executable= 47  executed=  0  never= 47  (0%)
> ```
>
> So the two functions from #377 are also untested — that is not this PR's doing, but it means the whole PSS feature currently has no coverage, and this PR is the natural point to fix that.
>
> There is a structural obstacle worth solving deliberately: because the salt is generated inside the function, the output is not reproducible, so it cannot be compared against the RFC 8017 test vectors. Two ways out, and I would suggest the second:
>
> 1. Compile-time override. `NX_CRYPTO_RAND` is overridable, so a test can build with `-DNX_CRYPTO_RAND=test_rand` and get a deterministic salt. This is exactly how I produced the byte-exact comparison in the summary, so I know it works — happy to hand over the harness, it is about 120 lines and already covers RSA-2048/3072/4096, the empty-PS boundary, the step 3 rejection and the scratch-size rejection.
> 2. Accept an optional salt. Add a `const UCHAR *salt, UINT salt_length` pair where `NX_CRYPTO_NULL` means "generate internally". That makes the function directly KAT-testable without build tricks, matches what most PSS APIs do, and would let a test use the published RFC 8017 vectors as-is.
>
> Either way, please also add a sign-then-verify round-trip test and a negative test (tampered EM must be rejected), and ideally one end-to-end TLS 1.3 client-certificate test with an RSA cert so the `nx_secure` side is covered too. Note that a test built on option 1 would also give #377's verify path its first coverage.

If you do this part, I could look into providing the rest of the test coverage for #377 myself.

nx_secure/src/nx_secure_tls_send_certificate_verify.c
```diff
@@ -528,6 +544,43 @@ NX_CRYPTO_EXTENDED_OUTPUT  extended_output;
...
+                 * at full size — the PKCS#1 v1.5 padding loop further down is gated to skip when nx_secure_tls_1_3 is set. */
+                current_buffer[length]     = (UCHAR)((tls_session -> nx_secure_tls_signature_algorithm) >> 8);
+                current_buffer[length + 1] = (UCHAR)((tls_session -> nx_secure_tls_signature_algorithm) & 0xFFu);
+                current_buffer[length + 2] = (UCHAR)(data_size >> 8);
+                current_buffer[length + 3] = (UCHAR)(data_size);
+                length += 4;
+
+                /* Sentinel: signature_length == data_size makes the PKCS#1 loop body below run zero iterations
+                 * (i < data_size - data_size - 1 underflows; we explicitly skip it via the nx_secure_tls_1_3 guard). */
+                signature_length = data_size;
```
fdesbiens: > The comment above this line says setting `signature_length == data_size` "makes the PKCS#1 loop body below run zero iterations (`i < data_size - data_size - 1` underflows)". That reasoning is inverted, and the conclusion is the opposite of the truth.
>
> `data_size`, `signature_length` and `i` are all `UINT`. With the two equal, `data_size - signature_length - 1` evaluates to `0xFFFFFFFF`, so `for (i = 2; i < 0xFFFFFFFF; ++i)` would run about four billion iterations writing `0xFF` across all of memory — not zero iterations. The only thing preventing that is the `if (!tls_session -> nx_secure_tls_1_3)` guard you added at `:668`.
>
> The guard is correct and the code is safe today. The problem is the comment: it tells the next reader that the underflow is harmless, which is precisely the belief under which someone would later delete the guard as redundant. That is a nasty trap to leave in a signing path.
>
> Please either drop the assignment entirely, since nothing downstream reads `signature_length` on the TLS 1.3 path, or set it to a meaningful value — and in both cases replace the comment with a plain statement that the PKCS#1 padding loop must not run for TLS 1.3, which is what the guard at `:668` enforces.

nx_secure/src/nx_secure_tls_process_certificate_request.c
```diff
@@ -120,12 +120,18 @@ UINT extension_type;
...
+            /* TLS 1.3 + RSA cert: announce RSASSA-PSS (RFC 8446 §4.2.3 — PKCS#1 v1.5 is banned for
+             * CertificateVerify in TLS 1.3). The send_certificate_verify path builds the EMSA-PSS-ENCODE
+             * via _nx_crypto_rsa_pss_sign. We pick rsa_pss_rsae_sha256 (0x0804) unconditionally — it's
+             * the only PSS sigalg our nx_crypto_rsa.c currently supports for signing, and any modern
+             * broker (mosquitto/OpenSSL) advertises it in CertificateRequest.sig_algs. */
             if (local_certificate -> nx_secure_x509_public_algorithm == NX_SECURE_TLS_X509_TYPE_RSA)
             {
-                return(NX_SECURE_TLS_UNSUPPORTED_CERT_SIGN_TYPE);
+                expected_cert_type = NX_SECURE_TLS_CERT_TYPE_RSA_SIGN;
+                expected_sign_alg = 0x0804u; /* rsa_pss_rsae_sha256 */
```
fdesbiens: > `0x0804`, `0x0805` and `0x0806` appear as bare literals here and again as `case` labels at `nx_secure_tls_send_certificate_verify.c:196`, `:199` and `:202`. Every other signature algorithm in this code has a name — `NX_SECURE_TLS_SIGNATURE_ECDSA_SHA256` and friends.
>
> There are currently no PSS constants anywhere in `nx_secure/inc/`, so please add them alongside the existing signature-algorithm defines, e.g. `NX_SECURE_TLS_SIGNATURE_RSA_PSS_RSAE_SHA256` and the 384/512 variants, and use them in both files. It also makes the eventual `rsa_pss_pss_*` extension easier to slot in.

nx_secure/src/nx_secure_tls_send_certificate_verify.c
```diff
@@ -23,13 +23,17 @@
...
+#include "nx_crypto_rsa.h"
 #ifdef NX_SECURE_ENABLE_DTLS
 #include "nx_secure_dtls.h"
 #endif /* NX_SECURE_ENABLE_DTLS */
 
 #ifndef NX_SECURE_DISABLE_X509
 static UCHAR handshake_hash[64 + 34 + 64]; /* We concatenate MD5 and SHA-1 hashes into this buffer, OR SHA-256/384/512. */
 static UCHAR _nx_secure_padded_signature[600];
+#if (NX_SECURE_TLS_TLS_1_3_ENABLED)
+static UCHAR _nx_secure_pss_scratch[600]; /* PSS encode: db[<=511 B] + salt[<=64 B] for RSA-4096+SHA-512 */
```
fdesbiens: > This brings the file to roughly 1.4 KB of static RAM — `handshake_hash[162]`, `_nx_secure_padded_signature[600]` and now this — which is a noticeable amount on the class of device ThreadX targets.
>
> The sizing comment says "db[<=511 B] + salt[<=64 B] for RSA-4096+SHA-512", which checks out: RSA-4096 with SHA-512 needs 447 + 64 = 511 bytes, so 600 is comfortable. But the buffer is only ever needed for the duration of one call, and the function already fails cleanly with `NX_CRYPTO_INVALID_BUFFER_SIZE` when the scratch is too small, so there is room to be tighter — 512 would do for everything up to RSA-4096, and deriving it from a documented maximum modulus constant would be clearer than a bare 600.
>
> Also worth stating in a comment: because this is `static`, two TLS sessions signing concurrently will corrupt each other. That is the existing pattern in this file rather than something you introduced, but this adds one more instance of it, and it is not obvious to a reader.

crypto_libraries/src/nx_crypto_rsa.c
fdesbiens: > Every other externally visible function in this file is declared `NX_CRYPTO_KEEP` — `_nx_crypto_rsa_operation` at `:82`, `_nx_crypto_method_rsa_init` at `:188`, `_nx_crypto_method_rsa_cleanup` at `:263`, `_nx_crypto_method_rsa_operation` at `:327`. Neither `_nx_crypto_rsa_pss_sign` nor `_nx_crypto_rsa_pss_verify` has it.
>
> That macro controls linker section placement and keep-out on several ports, so the omission can change what a port emits. #377 has the same gap, so fixing both here would be tidy.

nx_secure/src/nx_secure_tls_process_certificate_request.c
```diff
@@ -120,12 +120,18 @@ UINT extension_type;
...
             }
-
+            else
```
fdesbiens, line 138: > The new `else { ... }` wraps the existing curve `switch` but leaves its body at the old indentation, so the block's closing braces end up stacked and the `switch` looks as though it is outside the `else`. Please re-indent the wrapped lines one level.

crypto_libraries/src/nx_crypto_rsa.c
fdesbiens, line 794: > The new function's header block has `DESCRIPTION`, `INPUT` and `OUTPUT`, but omits `AUTHOR`, `CALLS`, `CALLED BY`. This should be added.
>
> Another detail to fix while you are in there. The `INPUT` text describes `scratch` as "must be >= db_len bytes", but the code requires `db_len + s_len` (`:823`) — the documented requirement is smaller than the real one, which is the wrong direction for a caller to be misled in.

fdesbiens (changes requested):
Thank you for this — it is careful work, and the parts that are hardest to get right are right. Before the findings, what I verified as correct, because it is worth recording:

**The PSS encoding is byte-exact.** I extracted `_nx_crypto_rsa_pss_sign()` into a harness, made the salt deterministic by compiling with `-DNX_CRYPTO_RAND=test_rand`, and compared the resulting EM against an independent EMSA-PSS-ENCODE implementation:

```
SHA-256 / RSA-2048         modulus= 256  emLen= 256  MATCH
SHA-256 / RSA-3072         modulus= 384  emLen= 384  MATCH
SHA-256 / RSA-4096         modulus= 512  emLen= 512  MATCH
SHA-256 / 66-byte mod      modulus=  66  emLen=  66  MATCH   <- emLen == hLen + sLen + 2, empty PS
EMSA-PSS-ENCODE (RFC 8017 §9.1.1) byte-exact agreement: ALL MATCH
```

Every step is correct: `emLen = ceil(emBits/8)`, the step 3 rejection, `DB = PS || 0x01 || salt` with the right PS length, MGF1 over H, the XOR, clearing the top `8*emLen - emBits` bits, and the `0xBC` trailer. The boundary case where PS is empty works, a 65-byte modulus is correctly refused, and an undersized scratch is correctly refused. Each EM round-trips through `_nx_crypto_rsa_pss_verify()`, and flipping one bit of mHash makes the verifier reject it. No ASan or UBSan findings on any of those runs.

**The TLS 1.3 wiring is correct.** Two things I specifically checked because they are the usual sources of interop failure:

- The value passed as mHash really is the right thing. By the time the PSS call is reached, `handshake_hash` holds the *digest* of `0x20 × 64 || context || 0x00 || transcript_hash` (built at `nx_secure_tls_send_certificate_verify.c:231-257`, hashed immediately after, with `handshake_hash_length` reset to the digest length). That matches RFC 8446 §4.4.3.
- The `0x0804` selection is not blindly trusted. `_nx_secure_tls_process_certificate_request()` still matches `expected_sign_alg` against the server's advertised `signature_algorithms` list and returns `NX_SECURE_TLS_UNSUPPORTED_CERT_SIGN_ALG` if the server does not offer it, and `nx_secure_tls_signature_algorithm` is only assigned on a match. So a server that does not advertise `rsa_pss_rsae_sha256` fails cleanly rather than receiving a signature it will reject. Good.

The `tls_1_3_enable_build_coverage` configuration builds clean under `-Werror -Wall -Wextra`, and all 150 nx_secure tests pass in that configuration, so nothing existing regresses.

Now the findings. One is blocking:

**`_nx_crypto_rsa_pss_sign()` writes `emLen` bytes into a buffer whose size it is never told.** The only caller passes the 600-byte `_nx_secure_padded_signature` with `em_bits` computed from the certificate's modulus length, which nx_secure bounds nowhere. With a 6144-bit RSA client certificate this walks off the end. Reproduced under ASan; details on `nx_crypto_rsa.c:789`.

**The salt does not go through `NX_CRYPTO_RBG`,** and the comment justifying that is incorrect — ECDSA does not use `NX_CRYPTO_RAND()` directly. Details on `nx_crypto_rsa.c:837`.

**There is no test coverage.** gcov on the coverage build reports 0 of 47 executable lines in `_nx_crypto_rsa_pss_sign`, and the same for the two functions #377 added. `AGENTS.md` asks for matching regression tests and 100% coverage. Details on `nx_crypto_rsa.c:760`, including a concrete way to make the function KAT-testable — I already have a harness that does it and am happy to hand it over.

Everything else is minor. Thank you also for basing this on `dev`.

fdesbiens:
Thank you for your meaningful contribution, @EdouardMALOT, and for your patience. I just submitted my review.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Fixed in dc4716e9, both parts. The function now takes an `em_length`, and the call site rejects `data_size > sizeof(_nx_secure_padded_signature)` before the version branch, so the PKCS#1 v1.5 path at :649 is covered too.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Fixed in 39bc5889 with the code you proposed, and the comment no longer claims parity with ECDSA. Builds clean with and without `NX_CRYPTO_SELF_TEST`.

Reading `_nx_crypto_drbg` I noticed it returns on a generate failure without releasing `NX_CRYPTO_DRBG_MUTEX_PUT`. Harmless by default since the macros are empty, but it leaks the mutex on ports that define them. It is unrelated to this PR so I kept it out — I will open a separate one.

nx_secure/src/nx_secure_tls_send_certificate_verify.c
EdouardMALOT: Fixed in 59faf341. You were right, and I took the first option: nothing reads `signature_length` on the TLS 1.3 path, so the assignment is gone.

One correction to the note: the guard is still needed, but `signature_length` is initialised to 0 at :95, so without it the loop bound is `data_size - 1` and it overwrites the PSS EM with 0xFF rather than running four billion iterations. The comment says that now.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Done in 5507ddab, and the `scratch` line was fixed in dc4716e9. I did the same for `_nx_crypto_rsa_pss_verify` and `_nx_crypto_rsa_pss_mgf1`, since they had the same gaps.

EdouardMALOT:
Thanks for the review. Pushed four commits: the blocking finding, the two other substantive ones, and the header blocks. The remaining minor items and the test coverage are still to come.

nx_secure/src/nx_secure_tls_process_certificate_request.c
EdouardMALOT: Done in 5826b5c1. Added `NX_SECURE_TLS_SIGNATURE_RSA_PSS_RSAE_SHA256/384/512` next to the existing signature-algorithm defines and used them in both files.

I documented them as opaque SignatureScheme code points rather than hash/signature pairs, since unlike the packed values above they are not composed that way, and noted where the `rsa_pss_pss_*` variants would slot in.

nx_secure/src/nx_secure_tls_process_certificate_request.c
EdouardMALOT: Fixed in 5826b5c1.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Fixed in 69f4722d, on `_nx_crypto_rsa_pss_sign` and `_nx_crypto_rsa_pss_verify`.

I also added it to the static `_nx_crypto_rsa_pss_mgf1`: `nx_crypto_ccm.c` uses `NX_CRYPTO_KEEP static` for its file-local routines, so that appeared to be the convention. Happy to drop that third one if you would rather keep it to externally visible functions.

nx_secure/src/nx_secure_tls_send_certificate_verify.c
EdouardMALOT: Done in 5826b5c1. The buffer is now sized from `NX_SECURE_TLS_PSS_MAX_MODULUS_SIZE` (512, RSA-4096), saving 88 bytes, and the comment states the static-buffer hazard you pointed out.

One consequence worth flagging: 600 bytes of scratch would have carried a modulus up to roughly 4800 bits, whereas 512 stops at RSA-4096. Anything larger now fails with `NX_CRYPTO_INVALID_BUFFER_SIZE` instead of being signed. That looks like the right trade given the RAM, but say the word if you would rather keep the headroom.

I checked the worst case rather than trusting the arithmetic. A sign/verify round-trip harness gives:

```
SHA-256 / RSA-2048           mod=256  sign+verify -> MATCH
SHA-256 / RSA-4096           mod=512  sign+verify -> MATCH
SHA-512 / RSA-4096 (worst)   mod=512  sign+verify -> MATCH     <- 447 + 64 = 511 of 512 used
SHA-512 / 130-byte mod       mod=130  sign+verify -> MATCH     <- emLen == hLen + sLen + 2, empty PS
SHA-512 / 129-byte mod       mod=129  refused, NX_CRYPTO_NOT_SUCCESSFUL
SHA-256 / 576-byte mod       mod=576  refused, NX_CRYPTO_INVALID_BUFFER_SIZE
```

A tampered mHash is rejected in every MATCH case.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Yes please, I would gladly take the harness. The deterministic-salt mechanism is exactly the part I did not want to invent twice, and starting from yours means the tests will match what you expect.

I have a local sign/verify round-trip harness covering the boundaries (worst-case scratch, empty PS, undersized modulus, tampered mHash), but it is a standalone program rather than a regression test, and a round-trip alone does not pin the encoding the way a KAT does.

My plan, unless you would rather structure it differently: fold your KAT vectors plus those boundary cases into `test/regression/crypto_test`, covering `_nx_crypto_rsa_pss_sign`, `_nx_crypto_rsa_pss_verify` and `_nx_crypto_rsa_pss_mgf1`, so the two functions from #377 get covered at the same time.

EdouardMALOT:
Pushed 69f4722d and 5826b5c1, covering the four remaining minor items: the named PSS scheme constants, the `NX_CRYPTO_KEEP` attributes, the scratch buffer sizing, and the indentation. I resolved the threads that are now purely done and left open the two where I asked you a question (the RSA-4096 ceiling that comes with the smaller scratch, and whether `NX_CRYPTO_KEEP` belongs on the static `_nx_crypto_rsa_pss_mgf1`).

The four files build clean under `-Werror -Wall -Wextra -Wconversion -Wsign-conversion` in four configurations, including `NX_CRYPTO_SELF_TEST`.

That leaves test coverage as the one substantive item outstanding, and I have taken you up on the harness offer in that thread. I will follow up with the regression tests.

EdouardMALOT:
Rebased onto current `dev` (force-push, `5826b5c1` -> `6d7ed660`). No conflicts, and the seven commits are unchanged.

Worth doing rather than cosmetic: `nx_secure_tls_send_certificate_verify.c` has moved under this branch since #417, so the merge result had never actually been compiled anywhere. It builds clean now under `-Werror -Wall -Wextra -Wconversion -Wsign-conversion` in four configurations, and the PSS sign/verify round-trip still passes on the rebased tree, including the RSA-4096 + SHA-512 worst case.

Apologies for any review anchors this moves. Doing it now rather than later keeps it to a single force-push before the coverage work lands.

EdouardMALOT:
rebased on "dev"

fdesbiens:
Hi Edouard — I rebased this branch onto the latest dev, removed the RSA-PSS definitions that are now already present
there, and added the required AI-disclosure comments. I did not change the functional implementation.

I have been working heavily on our CI workflows this week; we'll be in top shape for the September release.

The remaining items are regression tests and documentation. When do you expect to have those ready?

For the documentation, please update the “TLS 1.3 and X.509 Updates” section in
rtos-docs/netx-duo/modules/ROOT/pages/netx-duo-secure-tls/chapter2.adoc

(https://github.com/eclipse-threadx/rtos-docs-asciidoc/blob/main/rtos-docs/netx-duo/modules/ROOT/pages/netx-duo-secure-tls/chapter2.adoc#L59-L61)

and link the documentation PR here. The existing paragraph covers verification only; it should also document RSA-PSS
signing for TLS 1.3 RSA client-certificate authentication.

Regression tests still need to exercise the new RSA-PSS routines and the TLS 1.3 RSA client-authentication path with the
required coverage.

crypto_libraries/src/nx_crypto_rsa.c
EdouardMALOT: Done in 05f9d34: `nx_secure_rsa_pss_test.c` with its vectors, and `nx_secure_tls_1_3_rsa_client_auth_test.c` for the nx_secure side.

Coverage: 100% of `_nx_crypto_rsa_pss_verify` and `_nx_crypto_rsa_pss_mgf1`, so #377's functions are covered too and you need not spend time on that. 44 of 47 lines of `_nx_crypto_rsa_pss_sign` — the three left are the failure return after `NX_CRYPTO_RBG()`, your option 1: the macro is resolved at compile time, so closing it needs a build configuration that overrides it. Say the word and I will add one.

EdouardMALOT:
Both remaining items are done rather than scheduled, so no date needed.

Documentation: eclipse-threadx/rtos-docs-asciidoc#43.

Tests: 05f9d34, details in the thread on `nx_crypto_rsa.c`. One point worth making outside it: the nx_secure test drives the two functions directly rather than running a handshake, because `nx_secure_tls_send_certificate_request.c` builds the TLS 1.3 signature-algorithm list from the configured ECC curves — a NetX Duo server never offers `rsa_pss_rsae_sha256`, so a NetX-to-NetX handshake cannot reach this path. Extending the server side looks like a separate PR; happy to open one if you want it.

Fork CI: crypto 4/4 configurations green, nx_secure 19/19, no regression elsewhere.

fdesbiens:
Hi Edouard — heads-up that I pushed 0ff00235 straight to your branch rather than leaving it as review comments, since it was small and I wanted the test to prove itself. CI is green on it (16 pass, 3 skipped).

The PSS implementation itself checks out. I walked EMSA-PSS-ENCODE against RFC 8017 §9.1.1 step by step and it's correct, MGF1 can't overrun into H (it copies `min(hash_len, remaining)`), and the padding-loop guard is doing real work.

What I changed: `NX_SECURE_TLS_PSS_MAX_MODULUS_SIZE` sized the scratch buffer but nothing checked against it. Its comment credits the `em_length` check in `_nx_crypto_rsa_pss_sign`, but that one measures against `_nx_secure_padded_signature` (600 bytes), so the effective cutoff was 513 via the scratch check — and it surfaced `NX_CRYPTO_INVALID_BUFFER_SIZE` instead of a certificate error. Added the explicit check, fixed both comments, and added a 520-byte regression case. Negative control: without the check it fails with 0x20006 instead of 0x112. 154/154 in `tls_1_3_enable_build_coverage`.

Also dropped the `expected_cert_type` write in the TLS 1.3 RSA branch — TLS 1.3 has no `certificate_types` field, so it was dead — and recorded the `emBits = modBits - 1` assumption at the call site.

The static-buffer concurrency point you documented on `_nx_secure_pss_scratch` is now #430, assigned to me. Not yours to fix here.

One thing left before I'm comfortable merging: a third-party interop check. The round trip through our own `_nx_crypto_rsa_pss_verify` is good coverage, but it can't catch a systematic encoding error — both sides would have to be wrong the same way. Verifying one captured CertificateVerify with `openssl pkeyutl -verify -pkeyopt rsa_padding_mode:pss -pkeyopt rsa_pss_saltlen:-1` would close it. Happy to take that on if you'd rather not.

Extending the server side to offer `rsa_pss_rsae_sha256` as a separate PR sounds right to me — worth opening.

Docs PR eclipse-threadx/rtos-docs-asciidoc#43 is held until this lands. Your "4096 bits or less" line is exactly right now, which it wasn't before this fix.

EdouardMALOT:
@fdesbiens :
Done. `openssl pkeyutl -verify -pkeyopt rsa_padding_mode:pss -pkeyopt rsa_pss_saltlen:-1` accepts our signature on RSA-2048, 3072 and 4096. Flip one bit of the digest and it fails.

It isn't a wire capture though: the CertificateVerify content is built in a harness and signed through `_nx_crypto_rsa_pss_sign`. That still rules out a systematic encoding error. If you can grab a real one off the wire, that would cover the transcript content too.

Caveat on the harness: I plugged a reference SHA-256 into the `NX_CRYPTO_METHOD` and did the modexp outside NetX. It needs a 32-bit build (`-m32`, like CI) or `nx_crypto_sha2.c` returns wrong digests. Which accidentally made your point: with the bundled hash, sign and verify still agreed and the round trip passed. Both wrong the same way.

RSA-2048 case if you want to check it yourself:

```
-----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAx8lOgN+Vt99/FnaHAFOv
PZ5hkOGi4ckTwnWHIfuMlyv0ZjsKO5oFg2YsN05QoDdpSHhFZqoNwsUEVttjlCwB
J/WYKuttuBtRBkDs6atsIIire0/SK7CYWZIRrGavl1Be1fD/WTF+rq+sgqQF1tRR
9aSx+iVbkiOs+Sm1R6xttkGunib3pHzQ40SE4EidI0wOT5RfZFy7u/4/acSCp4n6
M4Zd6tVO5UIifM1nsjDRRNrLR7AFDj897Prvuzp8PmLqqNn4OY2rAHbEdPrczHem
sGY6aZ6duIYk1CyjfXN9qVkQP9sOBobA/cOyS1cXOc0wGMISKI8wIFcM52RvjD+q
zwIDAQAB
-----END PUBLIC KEY-----
```

```
digest.bin (sha256 of the signed content, hex):
81b01916a40b876d10f137d64e9798927ac490b33ede08f6bd8702a672bc4d39

sig.bin (base64):
itqnd6VrJNB4ZyUVPn8iyl/HfFZKJk+GJo9llN6vs91GN9B7VWH1kougQtWpUIxNrs89U8Ug/7/n
9xwKeePTd1KfK1J2npRshydSZygd7QJqB3s/HEGT6IVDW/6TmJfXj63qmn01xaaglDhOw51/qXfB
qhrCByCZCp12svc5M1FI4W4mfeD8fje0uPBPFoeYcT4Eil5wjcCGdBYcJiBS+BaRkwAtP24wx+0J
Uf/cNZw8Zqa8FDtFEDsjJX0yNR0KMuUWvTK26dJlCwrAn/qeAZKBy6zhCUUySE9I38pjsjzF9ozv
4HdEau+Tz1xiEA62Nl0ddNd11W7My5HWqZeyFg==
```

```
openssl pkeyutl -verify -pubin -inkey pub.pem -sigfile sig.bin -in digest.bin \
  -pkeyopt rsa_padding_mode:pss -pkeyopt digest:sha256 -pkeyopt rsa_pss_saltlen:-1
```

EdouardMALOT:
Corrected the comment in 9e31b6d3 following the finding on #431: the buffer is serialised by `_nx_secure_tls_protection`, which `receive_records` holds across record processing.

fdesbiens (approved):

crypto_libraries/src/nx_crypto_rsa.c
fdesbiens: Addressed in de26e4dc — both functions now carry NX_CRYPTO_KEEP, matching the rest of the file. Resolving.

PR #411, opened 2026-07-29
Fixed control frame handling in the WebSocket client

### LINKED ISSUE #368: WebSocket: Client does not respond with Pong after receiving Ping
**Describe the bug**
After the WebSocket connection is established and running normally, the server sends a Ping control frame, but the client does not respond with a Pong frame as required by the protocol. As a result, the server eventually closes the connection.

According to RFC 6455 (Section 5.5.2):

> Upon receipt of a Ping frame, an endpoint MUST send a Pong frame in response as soon as practical.

The current behavior is not compliant with the specification.

**To Reproduce**
1. Establish a WebSocket connection
2. Server sends a Ping frame
3. Client does not send Pong
4. Server closes the connection after timeout

**Expected behavior**
- Client receives Ping
- Client immediately sends Pong
- Pong payload matches Ping payload
- Connection remains alive

Yves57:
Fixes #368

A control frame carrying no payload never reached the frame switch: trimming its two-byte header consumed the whole packet, so the trim returned `NX_NO_PACKET` and an unconditional `return(NX_CONTINUE)` stopped short of it. A CLOSE never reached the status callback and a PING was never answered. Two more defects sat in the same paths. A control frame between the fragments of a message was rejected as an invalid termination frame and closed the connection, though RFC 6455 Section 5.4 permits it. And a 125-byte outgoing payload took the extended-length branch although only six header bytes were reserved, corrupting two payload bytes and declaring a length the wire did not match.

Every arm of the frame switch now handles the NULL packet, a PING is answered with a PONG echoing its payload, control frames are accepted between fragments without disturbing the message state, and 125 is encoded in the 7-bit field. A fragmented control frame, or one declaring more than 125 bytes, is rejected; that length check is the only bound on the copy into the PING echo buffer. The PONG is sent with the client mutex held throughout and with no unbounded wait, which is why the TLS packet allocate now honours its wait option.

14/14 websocket tests pass, three of them new. Each new case was checked to discriminate: without the empty-frame guards the test segfaults, restoring `< 125` fails the length assertion, and dropping the oversize check aborts with "stack smashing detected".

Assisted-by: Claude Code (Opus 5) <noreply@anthropic.com>


addons/websocket/nx_websocket_client.c
```diff
@@ -2115,6 +2125,74 @@ UCHAR *data_ptr;
...
+
+            status = _nx_packet_data_append(pong_packet, ping_payload, client_ptr -> nx_websocket_client_frame_data_length, client_ptr -> nx_websocket_client_packet_pool_ptr, NX_WAIT_FOREVER);
+            if (status != NX_SUCCESS)
+            {
+            	_nx_packet_release(pong_packet);
+                return(status);
+            }
+
+            /* Release the mutex */
+            tx_mutex_put(&(client_ptr -> nx_websocket_client_mutex));
```
Yves57: We release the mutex just because `_nx_websocket_client_send()` will unlock/lock it. Do we need to create a separate function `_nx_websocket_client_send_internal()` to avoid redundant lock/unlocks?
I'm not sure if it can lead or not a to race condition (like line ~2657) if a message is received exactly at this moment (necessarly from a different thread).

addons/websocket/nx_websocket_client.c
```diff
@@ -1727,6 +1728,8 @@ ULONG bytes_copied;
 ULONG packet_length;
 NX_PACKET *data_packet;
 UCHAR *data_ptr;
+UCHAR ping_payload[NX_WEBSOCKET_CONTROL_FRAME_PAYLOAD_MAXIMUM_LENGTH];
```
Yves57: To be moved in the `NX_WEBSOCKET_CLIENT` struct?

fdesbiens (changes requested):
Thanks Yves — the analysis behind this is right, and #368 is a real bug worth fixing. Your central observation is correct: when a control frame has no payload, trimming the header consumes the whole packet, `_nx_websocket_client_packet_trim()` returns `NX_NO_PACKET`, and the old unconditional `return(NX_CONTINUE)` meant the frame switch was never reached. CLOSE never reached the status callback and PING was never answered.

I have pushed two commits to your branch. **Please read the second half of this note before picking the work back up: the branch is currently red, and not only for the reason I fixed.**

## What I changed

`82477bc` — *Guarded the empty control frame paths in data_process*

Removing that early return also removed the only thing keeping `*packet_ptr == NX_NULL` out of the switch. `_nx_websocket_client_packet_trim()` returns `NX_NO_PACKET` **only** when it has released the packet and cleared the pointer — that is the `if (*packet_ptr) { ... return(NX_WEBSOCKET_SUCCESS); }` followed by `return(NX_NO_PACKET)` at the end of it. Your `if (*packet_ptr)` guard around the unmask block shows you had this in mind, but it needed to extend into the switch, where three of the four arms dereference the pointer:

- PING called `_nx_packet_data_extract_offset()`, whose first act is reading `nx_packet_length` through the pointer. It is the internal entry point, so there is no error-checking wrapper to catch the NULL.
- PONG called `_nx_websocket_client_packet_trim()`, whose first statement is the same kind of read.
- The data arms read `nx_packet_append_ptr`, so a zero-length TEXT or BINARY frame faulted too — and RFC 6455 permits those.

Only CLOSE survived, because it never touches the packet. That is almost certainly why the CLOSE behaviour you were chasing looked fixed while the rest was not.

A lone empty PING is exactly two bytes on the wire, so this is the ordinary keepalive shape rather than a corner case. It only appears intermittent because a following frame in the same segment leaves the pointer non-NULL and the fault does not occur.

Four other things in the same paths, in decreasing order of how much they mattered:

**The PONG send now keeps the mutex.** You released it around `_nx_websocket_client_send()`, which is correct in itself — that function acquires the mutex on entry, so it must be called unlocked. But by that point the trimmed packet had already been stored into `nx_websocket_client_processing_packet`, so during the window another thread calling `nx_websocket_client_disconnect()` could run `_nx_websocket_client_cleanup()` and release it. I moved the send into a helper that holds the mutex throughout instead: ThreadX mutexes are recursive, so `_nx_websocket_client_send()`'s nested acquire and release leave our ownership intact, and the window disappears rather than being narrowed.

**Every wait in that helper is `NX_NO_WAIT`.** Allocating with `NX_WAIT_FOREVER` while holding the client mutex deadlocks on an exhausted pool: the threads that would free packets need the same mutex. Note the other two allocation sites in the file pass the caller's `wait_option` rather than hardcoding a wait. Relatedly, failing to send a PONG no longer calls `_nx_websocket_client_cleanup()` — a lost PONG costs a keepalive round trip, which is not worth tearing down a working connection for.

**`bytes_copied` is now checked.** `_nx_packet_data_extract_offset()` reports a short copy through that parameter rather than through its status, so an unchecked result could echo uninitialised stack bytes from `ping_payload` to the peer.

**`NX_ASSERT` no longer stands in for a status check.** It expands to nothing in a default build (`nx_api.h:169`), so the trim result was being discarded; and when asserts are enabled, `NX_ASSERT_FAIL` is `for (;;) { tx_thread_sleep(NX_WAIT_FOREVER); }`, which hangs the thread rather than reporting anything. Either the condition cannot happen and the status should not be captured, or it can and needs handling — I went with handling it.

Also: the payload buffer is scoped to the PING case rather than the function, so the 125 bytes are not on the stack of every frame received; and the four tab characters are now spaces, since the file has none elsewhere.

`e85cf6e` — *Added a regression test for empty WebSocket frames*

`test/regression/websocket_test/netx_websocket_empty_control_frame_test.c`, registered in the netxduo regression CMake list. It sends an empty PING, PONG, data frame and CLOSE, each alone in its own TCP segment so the NULL case is genuinely reached, and checks the PING is answered with an empty PONG, that nothing is handed to the application as data, that the CLOSE reaches the callback and returns `NX_WEBSOCKET_DISCONNECTED`, that the mutex is released on every path, and that no packet is released twice.

I confirmed it discriminates rather than assuming it: on the parent commit it **segfaults**; with the fix it passes.

## The part that is still open, and why the branch is red

`netx_websocket_one_packet_with_multi_frames_test` **fails on this branch. It passes on `dev`.** It failed before my commits and it still fails after them — my changes neither caused nor fixed it.

This is the case you flagged in your PR description: *"in the case of interleaved frames (for example a PING frame received in the middle of several fragmented data frames), the next data frames are not correctly processed."* I would put it more strongly than the description does. That is not a pre-existing limitation left unfixed — it is a regression this PR introduces, and an existing test already catches it, as a segfault rather than a failed assertion.

The test's `server_response_frame_packet_1_3` is precisely that shape: the tail of a fragmented binary frame, then a PING with a two-byte payload, then a complete binary frame, all in one segment. The crash surfaces inside the test, at the point where it dereferences a packet after a non-zero status, but the cause is upstream: after the PING is handled, the frame that followed it is no longer delivered where the test expects it.

I instrumented both builds and compared the frame traces. The first ten frames are identical, PING and following binary frame included; `dev` then goes on to process four more. So the divergence is in sequencing after a control frame, not in the PING handling itself.

I stopped there rather than guessing. Getting it right means reworking how `nx_websocket_client_frame_fragmented`, `nx_websocket_client_frame_opcode` and `nx_websocket_client_frame_header_found` interact when a control frame arrives mid-fragmentation, and I would be inventing the intended semantics rather than restoring known-good behaviour. That is your call to make, and it is the work your PS was pointing at.

So: two of the three things in this PR are now done and covered by a test, and the third has gone from a note in the description to a failing test you can iterate against. If it would help to talk through the interleaving semantics before you change the state machine, I am happy to.

One practical note: I only added commits, so a plain `git pull` will bring them down — no reset needed.

Yves57:
Actually I don't think that the PR introduced a real regression in the test `server_response_frame_packet_1_3`.

This server does _really_ send a [PING](https://github.com/eclipse-threadx/netxduo/blob/master/test/regression/websocket_test/netx_websocket_one_packet_with_multi_frames_test.c#L104) during the test, and the new behavior (the expected PONG response) was creating a desynchronization of the frames. I fixed the test so that it waits for the PONG response before continuing the test.

Yves57:
An additional fix has been included in this PR: in addition to correctly handling the control frames in standard cases (i.e. outside of a fragmented frame sequence), control frames interleaved in a fragmented frame sequence are now also correctly processed.

Tested manually. And I have used an AI to generate a test case.

addons/websocket/nx_websocket_client.c
```diff
@@ -1430,7 +1432,7 @@ UINT header_size = NX_WEBSOCKET_HEADER_NORMAL_SIZE;
 
     /* Fill the next byte for MASK and Payload length.  */
     *data_ptr = NX_WEBSOCKET_MASK;
-    if (packet_ptr -> nx_packet_length < 125)
+    if (packet_ptr -> nx_packet_length <= 125)
```
Yves57, line 1435: Another bug fix (but important!): the header of messages with a payload of exactly 125 bytes was not correctly initialized.

Yves57:
Hi @fdesbiens,

Is there anything I can do to help with this MR?

fdesbiens:
You were right about `server_response_frame_packet_1_3` and I was wrong. Rather than take either of our words for it I built your branch at the test-only fix, before any of your product-side changes, and `netx_websocket_one_packet_with_multi_frames_test` passes five runs out of five. The failure was the test server never draining the PONG the client had begun sending, exactly as you described in August. It was not a sequencing regression and I have dropped that claim from the description.

The interleaving work then fixes a pre-existing defect rather than anything this PR introduced: before it, a PING arriving while `nx_websocket_client_frame_fragmented` was set took the termination-frame branch, hit `opcode != 0`, and closed the connection, which RFC 6455 Section 5.4 is explicit that you may not do.

I have pushed four commits, and one of them is worth your attention.

Your 125-byte limit on control frame payloads is load-bearing rather than conformance for its own sake. It is the only bound on the `_nx_packet_data_extract_offset()` copy into the 125-byte `ping_payload` buffer that the PING echo uses, so a peer declaring 126 or more runs off the stack. I removed that one check and the new test aborts with `*** stack smashing detected ***`. Nothing shipped, since the buffer and the guard both only ever existed on this branch, but it should not go in untested, so there is now a case that reaches it along with the fragmented-control-frame rejection and a 125-byte send.

The other three are smaller. `_nx_websocket_client_packet_allocate()` took a `wait_option` and passed `TX_WAIT_FOREVER` to the TLS allocate, so the `NX_NO_WAIT` the PONG path relies on to avoid blocking under the client mutex did nothing on a TLS connection. The AI disclosure comment on both files now matches the single fixed form the rest of the tree uses. And the three `frame_data_received < frame_data_length` guards in the CLOSE, PING and PONG arms are unreachable now that a control frame is only recorded once it is complete, so they are gone.

One thing I cannot do for you: you mentioned generating a test case with an AI. The Eclipse Foundation's generative AI guidelines want that recorded on the change, as `Assisted-by: <product> (<model and version>)`. Which tool and version was it? This will squash, so I can put it in the merge message next to mine and nothing needs rewriting on your side.

14/14 websocket tests pass locally on the current head. Once CI is green on it this is ready to merge from my side.

Yves57:
@fdesbiens 
> Which tool and version was it? This will squash, so I can put it in the merge message next to mine and nothing needs rewriting on your side.

The test 5 in `test/regression/websocket_test/netx_websocket_one_packet_with_multi_frames_test.c` was generated with Claude Opus 5. So the current header comment should be already correct.

fdesbiens (approved):

PR #419, opened 2026-08-13
Restored TLS 1.2 RSA signature fallback

fdesbiens:
## Description

TLS 1.3-capable clients replaced each configured RSA PKCS#1 signature scheme with its RSA-PSS counterpart. That satisfies TLS 1.3, but removes the RSA PKCS#1 schemes needed when the same ClientHello negotiates TLS 1.2. An OpenSSL TLS 1.2 server restricted to RSA/SHA-256 therefore reports that no suitable signature algorithm is shared.

This change:

- advertises `rsa_pss_rsae_sha256`, `rsa_pss_rsae_sha384`, and `rsa_pss_rsae_sha512` for TLS 1.3;
- follows each with the corresponding TLS 1.2 `rsa_pkcs1_sha256`, `rsa_pkcs1_sha384`, or `rsa_pkcs1_sha512` scheme;
- precomputes the exact encoded list size before writing, including overflow and undersized-buffer checks; and
- adds a focused regression test for wire order, TLS 1.2-only behavior, exact-fit and undersized buffers, unsupported entries, and the 16-bit extension-length limit.

No public API or documentation change is required.

## Testing

- CMake/Ninja `tls_1_3_enable_build_coverage` build
- Full TLS 1.3 regression suite: 153/153 passed
- Default non-TLS-1.3 configuration build and focused regression test passed
- New production paths and RSA mapping branches covered by the focused test
- `signature_algorithm_client_test` passed end to end against OpenSSL when combined in a disposable worktree with the iproute2 interoperability-test change; the runtime PATH deliberately contained no `ifconfig`


fdesbiens:
FYI @EdouardMALOT. I was working to fix the CI on this repository this week, and those changes were needed to pass the test suite.

Let me know what you think!

Issue #430, opened 2026-09-01
CertificateVerify signing uses static buffers shared across TLS sessions


fdesbiens:
## Summary

`_nx_secure_tls_send_certificate_verify` builds every CertificateVerify signature in file-scope `static` buffers. Two TLS sessions signing concurrently overwrite each other's working data, and the resulting signature is wrong for at least one of them.

## The buffers

All three are file-scope `static` in `nx_secure/src/nx_secure_tls_send_certificate_verify.c`:

| Buffer | Size | Role |
| :--- | :--- | :--- |
| `handshake_hash` | `64 + 34 + 64` | The RFC 8446 §4.4.3 signed content, then the digest of it |
| `_nx_secure_padded_signature` | 600 | The PKCS#1 v1.5 block, or the PSS encoded message |
| `_nx_secure_pss_scratch` | 512 | EMSA-PSS-ENCODE working area (`db` + `salt`) |

The first two predate RSA-PSS; the third arrives with #399. This is a pre-existing property of the design, not a regression — filing it so the constraint is recorded rather than rediscovered.

## Why the mutex does not cover it

`_nx_secure_tls_protection` is released before the handshake runs (`nx_secure_tls_session_start.c`), so two sessions can be inside `_nx_secure_tls_send_certificate_verify` at the same time. Nothing else serialises them.

## Consequences

A client authenticating on two TLS sessions at once can emit a CertificateVerify whose signature covers another session's transcript. The peer rejects it and the handshake fails, so the visible symptom is an intermittent authentication failure under concurrency rather than a security hole — the signature is invalid, not forgeable. It is timing-dependent and will not reproduce in single-session testing.

Server-side sessions are affected the same way: the function serves both roles, and a server signing CertificateVerify for two TLS 1.3 clients concurrently hits the same window.

## Options

1. **Move the buffers into `NX_SECURE_TLS_SESSION`.** Correct, and costs roughly 1.2 KB per session — significant on the class of device this targets.
2. **Allocate from the session's existing metadata area.** The crypto metadata regions are already per-session and sized at `nx_secure_tls_session_create` time, so this adds no new per-session field.
3. **Serialise `_nx_secure_tls_send_certificate_verify` behind its own mutex.** Cheapest, and correct given signing is not on a throughput path, but it blocks one handshake behind another.
4. **Document the constraint.** State that concurrent client-authenticating sessions are unsupported. Honest, but the failure mode is silent and intermittent.

[...]

PR #431, opened 2026-09-01
Moved the CertificateVerify buffers into the TLS session metadata

fdesbiens:
The CertificateVerify signature is built and checked in file-scope statics — `handshake_hash`, the PKCS#1 block (`_nx_secure_padded_signature` / `_nx_secure_decrypted_signature`) and the PSS scratch area. That state belongs to one session's transcript, so this moves it into the session rather than the translation unit.

The buffers are carved out of the application-supplied crypto metadata area at session create time. Sending and processing a CertificateVerify never overlap within one session, so the two paths share the same area: 764 bytes per session, or 1364 with TLS 1.3 enabled, and nothing at all when `NX_SECURE_DISABLE_X509` is defined.

## This is a trade, not a saving

Measured from the compiled objects, it removes **2660 bytes** of permanent `.bss` and adds **1364 bytes per session**:

| Sessions | static | per-session | delta |
| --- | --- | --- | --- |
| 1 | 2660 | 1364 | +1296 |
| 2 | 2660 | 2728 | −68 |
| 4 | 2660 | 5456 | −2796 |
| 8 | 2660 | 10912 | −8252 |

Break-even is just under two sessions. A single-session build gains about 1.3 KB; anything past two costs more than it saves. What the change buys is that these routines no longer depend on the global `_nx_secure_tls_protection` mutex for the integrity of their working state — that mutex is what serialises them today, and it is the only thing that does. Whether that independence is worth the memory is the question for review, and it turns on whether finer-grained TLS locking is wanted later.

`NX_SECURE_TLS_CERTIFICATE_VERIFY_SIGNATURE_SIZE` and `NX_SECURE_TLS_CERTIFICATE_VERIFY_PSS_SCRATCH_SIZE` let an application lower both when its largest certificate key is known to be smaller than the 4096-bit default.

## Breaking change

Because the required metadata grows, an application passing a hard-coded metadata buffer instead of the value from `nx_secure_tls_metadata_size_calculate` will get `NX_INVALID_PARAMETERS` from `nx_secure_tls_session_create` until it resizes the buffer.

## Also included

Both routines now reject a modulus larger than the signature buffer instead of writing past the end of it. That check was missing while the buffers were fixed 600-byte statics, and it matters more now that the size is configurable.

## Tests

`nx_secure_tls_certificate_verify_concurrency_test` pins the buffers as per-session state. It drives both routines directly and interleaves two sessions inside them, and it fails on the code before this change on both the signing and verifying sides. To be explicit about what it does and does not show: that interleaving is prevented through the public API by the protection mutex, so the test is asserting that each session keeps its own transcript, not that a reachable race exists. It also covers the two new modulus checks.

`nx_secure_tls_client_handshake_coverage_test` assembles a TLS session by hand rather than through `nx_secure_tls_session_create`, so it now supplies the scratch area the same way it already supplies the crypto table.

Rebased onto dev after #399; the RSA-PSS scratch added there is folded into the same per-session area, and the TLS 1.3 guard around the PKCS#1 padding loop is preserved. Full nx_secure suites pass: 155/155 in `tls_1_3_enable_build_coverage`, 179/179 in `default_build_coverage`, built with GCC 14.2.0.


fdesbiens:
@EdouardMALOT would you be willing to review this one? A formal review request is not possible because review requests are limited to repository collaborators, hence the ping.

You found this bug while working on #399, and that PR is the part of the tree most affected: the send path's PSS scratch will want to come from the per-session region this adds rather than a new static. The two review questions I would most value your eyes on are whether sharing one region between the send and process paths is safe in every handshake ordering you have exercised, and whether the growth in the required metadata size is acceptable for the targets you care about.

EdouardMALOT:
@fdesbiens Thanks for the ping. Both answers below, but the first one comes with a finding that I think changes the decision.

**1. Sharing one region between send and process.** Safe within a session: the two calls are sequential in the state machine, nothing is kept in the scratch between calls, and the signature is copied into the packet before return. But while checking every ordering I went back to the premise of #430, and it does not hold. `_nx_secure_tls_session_start` does release the mutex at line 175, but it then calls `_nx_secure_tls_handshake_process`, which goes through `_nx_secure_tls_session_receive_records`. That function takes `_nx_secure_tls_protection` again at line 90 before any record is processed, and `_nx_secure_tls_process_record` runs under it, so all four handshake state machines do too. DTLS does the same in `_nx_secure_dtls_session_receive` (lines 298-304), and renegotiation goes through `receive_records` as well. The mutex is only dropped around blocking calls (TCP send/receive, packet allocate, the transmit mutex, X.509 parse), none of which sit inside the two CertificateVerify routines.

So two sessions cannot be inside `_nx_secure_tls_send_certificate_verify` or `_nx_secure_tls_process_certificate_verify` at the same time through the public API. The new test reaches the overlap only because it calls the internal functions directly and preempts from a hooked hash method, which bypasses the mutex. It shows the buffers are shared, not that the race is reachable.

This is consistent with the rest of nx_secure: the same static pattern exists in about ten other handshake routines (`nx_secure_process_server_key_exchange.c`, `nx_secure_tls_ecc_generate_keys.c`, `nx_secure_x509_certificate_verify.c`, `nx_secure_generate_client_key_exchange.c`, `nx_secure_tls_process_finished.c`, `nx_secure_verify_mac.c`), all relying on the global mutex. Fixing two of them per-session leaves the design in between. Either the mutex is the design and #430 closes with a corrected comment on the statics, or the whole family moves to a generic per-session scratch region sized to the largest user, so the metadata does not grow again with each follow-up.

The wrong claim came from my own comment on `_nx_secure_pss_scratch` in #399. I will fix it there whatever happens to this PR.

**2. Metadata growth.** Not a concern for my target. ~1.4 KB per session is small in my case, and I would expect the same for most targets that run TLS 1.3. Two things worth stating in the docs if the change stays: a DTLS server multiplies it by the number of sessions, and `nx_secure_tls_certificate_verify_scratch_size` is written at create time but never read by the two routines, which use the compile-time constants.

If the region stays, #399 rebases as you describe: `pss_scratch` at `NX_SECURE_TLS_CERTIFICATE_VERIFY_PSS_OFFSET`, and the 512-byte check becomes redundant with your 600-byte one. The 520-byte regression case you added in 0ff00235 depends on the 512 check, so we would need to settle which bound is the documented one.

PR #435, opened 2026-09-15
Fix TCP low-watermark receive-window underflow and header corruption

### LINKED ISSUE #433: TCP: low-watermark drop path underflows `rx_window_current`, and the wrapped window overwrites the header flags (segments sent with RST set)
**DISCLAIMER:** this bug was found by AI while debugging connection resets on a custom PCB when trying to bridge UART over a TCP connection. The content of this issue has been AI-summarised. 

**Describe the bug**

When `NX_ENABLE_LOW_WATERMARK` is defined, `_nx_tcp_socket_state_data_check()` can take one segment through both reassembly (advancing `rx_sequence`) and the drop-tail branch. When it does, it decrements the receive window twice for the same advance, and `ULONG nx_tcp_socket_rx_window_current` wraps. Without `NX_ENABLE_TCP_WINDOW_SCALING`, the three TCP header builders OR that value into `nx_tcp_header_word_3` unmasked. The upper bits of the wrapped window then overwrite the data offset and flags, so every following segment leaves with **data offset 15 and all eight flags set, RST included**. The sequence number is correct and the checksum is valid, so the peer accepts the RST and the connection dies.

All line references are to `master` at 473d192.

1. **Trigger.** Segments are queued out of order, and the receive queue is at `nx_tcp_socket_receive_queue_maximum` (or the pool is below its low watermark). A segment then arrives that fills the hole. `drop_packet` is set ([nx_tcp_socket_state_data_check.c#L486](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_state_data_check.c#L486)).
2. **Reassembly.** The contiguity loop advances `rx_sequence` over the newly contiguous data. In the queue-full case it stops one packet before the tail ([#L926](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_state_data_check.c#L926)).
3. **Drop-tail branch.** The branch releases the tail and sets the window **absolutely** from the already-advanced `rx_sequence` ([#L968-L972](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_state_data_check.c#L968-L972)):
   ```c
   rx_window_current = tail_seq + tail_len - rx_sequence;   /* typically 0 */
   rx_window_last_sent = rx_window_current;
   ```
4. **Second decrement.** The generic update then subtracts the same advance **again** ([#L1009-L1017](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_state_data_check.c#L1009-L1017)):
   ```c
   if (original_rx_sequence != rx_sequence)
   {
       rx_window_current   -= (rx_sequence - original_rx_sequence);   /* 0 - 7624 = 0xFFFFE238 */
       rx_window_last_sent -= (rx_sequence - original_rx_sequence);
   }
   ```
5. **Header corruption.** Without window scaling, `window_size = rx_window_current` is ORed straight into `nx_tcp_header_word_3` by all three header builders:
   - [nx_tcp_packet_send_control.c#L216](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_packet_send_control.c#L216)
   - [nx_tcp_socket_send_internal.c#L728](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_send_internal.c#L728)
   - [nx_tcp_socket_retransmit.c#L283](https://github.com/eclipse-threadx/netxduo/blob/473d192822babbd831dc5907b9e279b340a3925b/common/src/nx_tcp_socket_retransmit.c#L283)

   The `> 0xFFFF` clamp only exists under `NX_ENABLE_TCP_WINDOW_SCALING`. With scaling enabled, the flags survive but the advertised window is still wrong (0xFFFF).

**A related defect in the same branch (silent data loss).** The loop's early exit (#L926) only protects the tail when the queue holds exactly `maximum + 1` packets. If the **pool low watermark** triggers the drop while the queue is short, or a superset retransmission shrinks the queue, the loop marks the tail `NX_PACKET_READY` and advances `rx_sequence` past it. The drop branch then releases it. Those bytes are ACKed to the peer but never delivered to the application. The window formula in step 3 also goes negative in this case (new tail end < `rx_sequence`), which is a second source of the wrap.

Environment:
- **Target:** STM32H563 (Cortex-M33) on a custom board. Ethernet MAC checksum offload is enabled, so the corrupted segments carry a valid checksum.
- **Version:** NetX Duo 6.4.0 as bundled in STM32CubeH5 1.7.0. With comments stripped, the code in the four affected files is identical on `master` (473d192, v6.5.1.202602_rel), and the patch below applies to `master` cleanly.
- **Toolchain:** Arm GNU Toolchain 14.2.Rel1 (arm-none-eabi-gcc 14.2.1), CMake, Release (-Os) and Debug builds.
- **Relevant configuration:** `NX_ENABLE_LOW_WATERMARK`, `NX_TCP_MAX_OUT_OF_ORDER_PACKETS 8`, `NX_TCP_IMMEDIATE_ACK`, `NX_ENABLE_TCP_WINDOW_SCALING` **not** defined. The sockets use `nx_tcp_socket_receive_queue_max_set(socket, 8)`, an 8192-byte window, and `nx_packet_pool_low_watermark_set()` on the default pool.
- **Peer:** Linux (Raspberry Pi 5).

Diagnosis:
- **Wire capture on the Linux peer (raw socket):** the capture shows the 244-byte retransmission that filled the hole. Every segment that followed carried flags `FIN SYN RST PSH ACK URG ECE CWR`, with window fields 57912, 58072, 59776 and 64236. As signed 16-bit values those are −7624, −7464, −5760 and −1300.
- **Live memory read on the target (non-halting SWD):** `rx_window_current == (ULONG)-7624` and `receive_queue_count == 8`, while `rx_sequence` had advanced by exactly 7624.
- **Breakpoint:** the target's own socket reset came in through NetX's received-RST path. That was the peer resetting the corrupted connection.
- **Host test:** a unit test compiles the real `nx_tcp_socket_state_data_check.c`, `nx_tcp_packet_send_control.c`, `nx_tcp_socket_send_internal.c` and `nx_tcp_socket_retransmit.c` against minimal fakes. It reproduces the corrupted header on the unmodified sources for the queue-cap case and both pool-watermark cases.

**To Reproduce**

1. Build with `NX_ENABLE_LOW_WATERMARK` and without `NX_ENABLE_TCP_WINDOW_SCALING`.
2. On an established socket, call `nx_tcp_socket_receive_queue_max_set(socket, 8)`. The window is 8192.
3. From the peer, with `rx_sequence = R`, send 8 data segments that start at `R + 244` (for example 980 bytes each). They are queued out of order; the queue count is 8.
4. Send the 244-byte segment at `R`.
5. Observe `rx_window_current` wrap to `0 - (244 + 7 * 980)`. Observe that the next ACK or data segment has a TCP header word of `0xFFFFxxxx` (data offset 15, all flags set).

Under real traffic we hit this at 1 Mbaud, bridging full-duplex UART over TCP to a Linux host with modest loss. In a 30-minute run, 6 of 47 connections were reset.

**Expected behavior**

- The window is decremented once per advance of `rx_sequence`. After a drop, it closes exactly at the end of the retained data.
- A dropped tail is never acknowledged before it is released.
- The advertised window cannot overwrite the data offset or flags, whatever `rx_window_current` holds.

**Impact**

A showstopper for us. Connections under load were reset at random, and in the pool-watermark case received data was silently lost. We now ship the local patch below.

**Logs and console output**

The wire and memory observations are summarised above. 

**Additional context**

Our patch against `master` (4 files, +44/−15) is attached as `netxduo-rx-window-upstream.patch`:

- **`nx_tcp_socket_state_data_check.c`, second decrement:** skip the relative window update when `drop_packet` is set, because the drop branch has already set the window absolutely. `original_rx_sequence` is left alone, so the later ACK decision is unchanged.
- **`nx_tcp_socket_state_data_check.c`, tail stop:** under `NX_ENABLE_LOW_WATERMARK`, stop the contiguity loop *before* the tail when `drop_packet` is set, so the tail is never acknowledged. In the queue-full case the stop point is unchanged. If a superset retransmission leaves only itself queued under pool pressure, it is dropped without an ACK and the peer retransmits it.
- **The three header builders:** a negative `rx_window_current` advertises 0, and the `> 0xFFFF` saturation applies in both window-scaling configurations.

The header-builder change is defensive: it stops any future window bug from turning into a corrupted header.

**patch:** [netxduo-rx-window-upstream.patch](https://github.com/user-attachments/files/32194761/netxduo-rx-window-upstream.patch)

Sawii00:
## Summary

Fixes #433.

When a segment fills a receive-queue hole under low-watermark pressure, the drop path can set the receive window using the advanced receive sequence and then subtract that advance again. The resulting wrap can corrupt the TCP flags, including setting RST. Reassembly can also acknowledge the tail before dropping it when pool pressure or a superset retransmission leaves a shorter queue.

This PR applies the four-file patch from the issue to `dev`:

- Stop reassembly before the tail that will be released.
- Skip the relative window decrement after the drop path has set the window absolutely.
- Advertise zero for a wrapped internal window and saturate oversized advertised windows in the control-send, new-data-send, and retransmission header builders, with or without window scaling.

## Regression coverage

Add `netx_tcp_low_watermark_rx_window_test` to the existing CMake regression suite. It uses real NetX sockets, packet allocation/release, TCP reassembly, send, and retransmission code. A sink IP driver inspects outgoing TCP headers; received segments are injected at the reassembly boundary to make the sequences deterministic.

Coverage includes:

- Queue-limit hole fill, plus a superset retransmission that shrinks a full queue.
- Pool-low-watermark drop with a short queue.
- Pool-low-watermark superset retransmission where the incoming packet becomes the tail.
- Receive sequence/window, ACK number, delivered payload bytes, queue cleanup, and packet-pool accounting.
- Ordinary, wrapped, and oversized windows through all three header builders, checking the exact data offset and flags. The scaling-enabled test uses a nonzero receive scale of 2.

Enable `NX_ENABLE_LOW_WATERMARK` in the existing `v4_build` and `v4_full_build` profiles. This runs the regression, along with existing low-watermark tests, without and with `NX_ENABLE_TCP_WINDOW_SCALING` respectively.

## Validation

Locally tested in Ubuntu 24.04, Linux x86 32-bit, GCC 14, with the repository-pinned ThreadX and FileX submodules:

- 46 selected TCP/low-watermark CTests pass in each profile, including overlap, wrapping sequences, retransmission, dropped packets, data transfer, and window handling.
- The new regression executes all scenarios successfully in both profiles; it is not N/A in either.
- The new test also passes strict C99 warning checks, including `-Werror -Wall -Wextra -pedantic -Wconversion -Wmissing-declarations`.
- Five local mutation checks fail as expected: restoring each of the four original source files individually, and removing only the new tail stop. Restoring the complete patch passes again.
- Profile-wrapper validation and `git diff --check` pass.

The selected CTests were run with:

```sh
ctest --test-dir test/cmake/netxduo/build/<profile> \
  -R 'netx_(low_watermark|tcp_.*(window|overlapping|wrapping_sequence|retransmi|dropped_packet|data_transfer))' \
  -E ipv6 -j 8 --timeout 180 --output-on-failure
```

The full repository test matrix has not been run locally.

## AI disclosure

The fix and regression coverage were developed with OpenAI Codex assistance. Edited files include AI disclosures, and the commit includes an `Assisted-by: OpenAI Codex` trailer. The Eclipse Contributor Agreement has been signed.


fdesbiens:
Thanks again, @Sawii00. The fix is correct and the regression is solid: both `v4_build` and `v4_full_build` pass in full, and reverting any one of the five changes makes the new test fail in both profiles.

I pushed one commit on top of yours that normalises the AI disclosures. The project recently settled on new rules for these. They are not public yet, but they will go out with next week's release. In short: an edited source file carries a single fixed line, `// Portions of this file were generated with AI assistance.`, under its header. A new file carries the standard header, whose disclosure names the product and model. Build files carry no disclosure. The per-change record is the `Assisted-by` trailer on the commit. No code changed, and I also dropped the issue number from the test's comment and banner.

I filled in the model as `gpt-5.6-sol` in the new test's header. If Codex reported a different model string for this work, please tell me and I will correct it. When I squash-merge, the trailer will read `Assisted-by: Codex (gpt-5.6-sol) <noreply@openai.com>`.

Two things I noticed during review are outside this PR's scope and are tracked in #444: the low-watermark code is not measured by any coverage profile, and the TCP_INFO receive counters still count dropped packets.

PR #438, opened 2026-09-18
Released the packet when the MQTT client receives a malformed PUBLISH

EdouardMALOT:
@fdesbiens — just found this one; flagging it in case it can still make the upcoming release.

## Summary

The MQTT client leaks one packet per **malformed PUBLISH** it receives.

`_nxd_mqtt_process_publish()` returns `NX_TRUE`, `NX_FALSE` or `NXD_MQTT_INVALID_PACKET` (0x1000F) when the PUBLISH is malformed — remaining length below 2, topic length past the end of the message, or QoS 1/2 with no room for the packet identifier (`nxd_mqtt_client.c:1287-1316`). `_nxd_mqtt_packet_receive_process()` stores that return value straight into `packet_consumed`, so the error code counts as "consumed": the final `nx_packet_release()` is skipped and nothing else references the packet.

The leaked packet is the one the driver received into, with or without TLS, so it comes from the driver's receive pool — usually the IP default pool. A malicious broker, or on a plain connection anyone able to inject into the TCP stream, can drain it with a few 3-byte messages and stop all inbound traffic until reboot.

## Changes

- **`addons/mqtt/nxd_mqtt_client.c`** — only set `packet_consumed` when `_nxd_mqtt_process_publish()` returns `NX_TRUE`. On `NXD_MQTT_INVALID_PACKET`, stop parsing the rest of the chain (new `packet_invalid` flag, as `break` in the `switch` cannot leave the `while`) and let the existing release path free the packet. Also listed the third return value in the function's `OUTPUT` block.

  Stopping at the malformed message rather than skipping it is deliberate: MQTT 3.1.1 §4.8 asks a client to close the connection on a protocol error, and dropping the rest of the segment is the closest this path gets without adding a disconnect.

- **`test/regression/mqtt_test/netx_mqtt_malformed_publish_test.c`** — regression test. The broker sends one PUBLISH per malformed case, then a well-formed one; the test asserts that only the latter is delivered, that the client keeps receiving, and that the pool is whole once the client is deleted. Plain TCP and, with `NX_SECURE_ENABLE`, TLS. These returns had no coverage before.

## Test plan

- [x] MQTT regression suite: 36/36 in all nine configurations
- [x] Test commit alone on `dev`: fails on the pool check, so the test discriminates
- [x] Well-formed PUBLISH handling unchanged — `NX_TRUE`/`NX_FALSE` take the same path as before


fdesbiens:
Thanks, @EdouardMALOT. We're continuing this change in another repository, and I've sent you the details there. Closing this one.

